Issue #12, opened 2022-10-13
Bringing URDNA2015 to FPWD


gkellogg:
In the [last meeting](https://www.w3.org/2022/10/12-rch-minutes.html,text), we discussed bringing [URDNA2015](https://w3c-ccg.github.io/rdf-dataset-canonicalization/spec/index.html)  to FPWD. Just to clarify:

* [x] This will be the FPWD of RDF-RDC, and not another publication.
* [x] The CCG has yet to act on publishing this spec as a Final CG Report, which was discussed at TPAC (nudge nudge).
  * [ ] Are we importing just the (eventually) published report in a single commit, or are we expecting to inherit its commit history?
* [x] Before we can publish, we need to flesh out examples and narrative material to make it better understandable. (See #1)
* [ ] A future version may replace URDNA2015 with another algorithm, which may or may not be derivative. (See #10)
* [x] The CG Draft includes a [test suite](https://w3c-ccg.github.io/rdf-dataset-canonicalization/tests/), which can presumably be re-purposed as the basis for our own test suite.
* [x] The CG Draft includes an appendix on URGNA2012, which I presume will continue on in the FPWD.
  * [x] Does this include [tests in the test suite](https://w3c-ccg.github.io/rdf-dataset-canonicalization/tests/#manifest-urgna2012)?
* [x] The Draft includes an Editor’s Note suggesting that the use of the _simple_ variable is unused and the algorithm should be updated accordingly. Presumably this, and other notes within the draft are migrated to GitHub Issues and not necessarily addressed for FPWD.

Once we settle some or all of these issues, I can begin work on this.

peacekeeper:
CCG has by now published a Final Community Group Report:
- https://www.w3.org/community/credentials/
- https://www.w3.org/community/reports/credentials/CG-FINAL-rdf-dataset-canonicalization-20221009/

gkellogg:
Ahh, interesting that they did not publish the version from https://github.com/w3c-ccg/rdf-dataset-canonicalization, which actually moved it from the json-ld repo to the w3c 
-ccg repo as we discussed.

gkellogg:
Or, perhaps they just chose not to merge https://github.com/w3c-ccg/rdf-dataset-canonicalization/pull/28 which made the links and names consistent with the location.

peacekeeper:
I created this PR now to update the FCGR: https://github.com/w3c/cg-reports/pull/16

gkellogg:
Closed via #29.

PR #13, opened 2022-10-14
Initial Import CG report

gkellogg:
This is based on the [CCG Final Report](https://www.w3.org/community/reports/credentials/CG-FINAL-rdf-dataset-canonicalization-20221009/) with modifications to bring it into the WG.

* Original ReSpec source from https://github.com/w3c-ccg/rdf-dataset-canonicalization/blob/main/spec/index.html.
* Add .pr-preview.
* Update ReSpec config and create bibliographic definitions.
* Add GitHub Actions CI step to validate spec using ReSpec action.

For #12.

(I would have added @dlongley as a reviewer, but he hasn't been added to the repo as of yet).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,25 @@
+# This workflow validates the document for markup and examples.
+name: CI
+
+on:
+  push:
+    branches: [ '**' ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  tests:
+    name: Build and Validate Spec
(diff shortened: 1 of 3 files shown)


gkellogg:
Note, if we decide to incorporate more GitHub history, as suggested in #12, we can re-base this work easily enough. Personally, I think the reference to the CG report is adequate for finding the history, if necessary.

dlongley (approved):
LGTM. Thanks, @gkellogg!

PR #17, opened 2022-10-17
Descriptive content

gkellogg:
* Update editors and authors, using w3cid. Adds @gkellogg, removes @msporny.
* Normalize spec references.
* Reference specific issues with some issues in the spec.
* Move localBiblio to a common file.
* Add typographical conventions.
* Add explicit section identifiers, updating some references.
* Minor intro updates.
* Add some descriptive information from the Arnold/Longley paper.

cc/ @dlongley

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rch-rdc/pull/17.html" title="Last updated on Oct 31, 2022, 3:59 PM UTC (e4f5846)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rch-rdc/17/c2ec31c...e4f5846.html" title="Last updated on Oct 31, 2022, 3:59 PM UTC (e4f5846)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -889,7 +889,7 @@
       <ol>
         <li id="ndh-hl-s1"><strong>Compute related hashes</strong>.
           Compute the related hash <var>H<sub>n</sub></var> set for <var>n</var>,
-          i.e. all first degree <a>mentions</a> between <var>n</var> and another blank node.
+          i.e., all first degree <a>mentions</a> between <var>n</var> and another blank node.
           Note that this includes both labeled and unlabeled blank nodes.</li>
         <li id="ndh-hl-s2"><strong>Explore mentions</strong>.
           Given the related hash <var>x</var> in <var>H<sub>n</sub></var>,
--- a/spec/index.html
+++ b/spec/index.html
@@ -190,9 +190,9 @@
     problematic dataset and may choose to assume it's an attempted denial of
     service attack, rather than a real input, and abort.</p>
 
-  <p>This document outlines an algorithm for generating a normalized
-    <a>RDF dataset</a> given an <a>RDF dataset</a> as input. The
-    algorithm is called the
+  <p>This document outlines an algorithm for generating a canonical
+    serialization of an <a>RDF dataset</a> given an <a>RDF dataset</a> as input.
+    The algorithm is called the
     <strong>Universal RDF Dataset Canonicalization Algorithm 2015</strong> or
     <a>URDNA2015</a>.</p>
 
@@ -373,7 +373,7 @@
       <dd>An <a>identifier issuer</a>, initialized with the
         prefix <code>_:c14n</code>, for issuing canonical
         <a>blank node identifiers</a>.
-        <div class="issue">
+        <div class="ednote">
           Mapping all <a>blank nodes</a> to use this
           identifier spec means that an <a>RDF dataset</a> composed of two
           different <a>RDF graphs</a> will use different
@@ -459,7 +459,12 @@
               <a>blank node identifier</a> in the
               <a>blank node to quads map</a>, creating a new entry if
               necessary.
-              <div class="issue">It seems that quads must be normalized, so that literals with different syntactic representations but the same semantic representations are merged, and that two graphs differing in the syntactic representation of a literal will produce the same set of blank node identifiers.</div>
+              <div class="issue" data-number="15">
+                It seems that quads must be normalized,
+                so that literals with different syntactic representations
+                but the same semantic representations are merged,
+                and that two graphs differing in the syntactic representation
+                of a literal will produce the same set of blank node identifiers.</div>
             </li>
           </ol>
         </li>
@@ -645,7 +650,10 @@
                   </ol>
                 </li>
               </ol>
-              <p class="issue">Note potential need to normalize literals to their canonical representation here as well, if not done on the original <a>input dataset</a>.</p>
+              <p class="issue" data-number="15">
+                Note potential need to normalize literals to their
+                canonical representation here as well,
+                if not done on the original <a>input dataset</a>.</p>
             </li>
           </ol>
         </li>
@@ -737,7 +745,7 @@
     <section>
       <h3>Algorithm</h3>
 
-      <div class="issue">
+      <div class="issue" data-number="16">
         An additional input to this algorithm should be added that
         allows it to be optionally skipped and throw an error if any
         equivalent related hashes were produced that must be permuted
--- a/spec/index.html
+++ b/spec/index.html
@@ -130,7 +130,7 @@
   <p>When data scientists discuss canonicalization,
     they do so in the context of achieving a particular set of goals.
     Since the same information may sometimes be expressed in a variety of different ways,
-    it often becomes necessary to be able to transform each of these
+    it often becomes necessary to transform each of these
     different ways into a single, standard representation.
     With a standard representation, the differences between
     two different sets of data can be easily determined,
@@ -178,33 +178,33 @@
       <li>when a node is used to stitch together parts of a graph and the nodes themselves are not interesting (e.g., <a data-cite="RDF11-MT#rdf-collections">RDF Collections</a> in [[RDF11-MT]]),</li>
       <li>when someone is trying to create an intentionally difficult graph topology.</li>
     </ul>
-    <p>Further more,
-      RDF semantics dictate that deserializing the same RDF document
+    <p>Furthermore,
+      RDF semantics dictate that deserializing an RDF document
       results in the creation of unique <a>blank nodes</a>,
       unless it can be determined that on each occasion,
-      the <a>blank node</a> identifies the same resource;
-      this is due to the fact that <a>blank node identifiers</a>
+      the <a>blank node</a> identifies the same resource.
+      This is due to the fact that <a>blank node identifiers</a>
       are an aspect of a concrete RDF syntax
       and are not intended to be persistent or portable.
       Within the abstract RDF model,
       blank nodes do not have identifiers
       (although some
       <a data-cite="RDF11-CONCEPTS#dfn-rdf-store">RDF store</a>
-      implementations may use stable identifiers and choose to make them portable).
+      implementations may use stable identifiers and may choose to make them portable).
       See <a data-cite="RDF11-CONCEPTS#section-blank-nodes">Blank Nodes</a>
       in [[!RDF11-CONCEPTS]] for more information.</p>
 
     <p>RDF does have a provision for allowing blank nodes
       to be published in an externally identifiable way through the use of
       <a data-cite="RDF11-CONCEPTS#dfn-skolem-iri">Skolem IRIs</a>,
-      which allows a given RDF store to replace the use of blank nodes
+      which allow a given RDF store to replace the use of blank nodes
       in a concrete syntax with <a>IRIs</a>,
-      which serve to repeatably identify that blank node within that particular RDF store,
+      which then serve to repeatably identify that blank node within that particular RDF store;
       however, this is not generally useful for talking about the
       same graph in different RDF stores,
       or other concrete representations.
       In any case, a stable <a>blank node identifier</a> defined for one
-      RDF store, serialization is arbitrary,
+      RDF store or serialization is arbitrary,
       and typically not relatable to the context within which it is used.</p>
 
     <p>This specification defines an algorithm for creating stable <a>blank node identifiers</a> repeatably for different serializations possibly using individualized <a>blank node identifiers</a> of the same RDF graph (dataset) by grounding each <a>blank node</a> through the nodes to which it is connected, essentially creating <em>Skolem <a>blank node identifiers</a></em>. As a result, a graph signature can be obtained by hashing a canonical serialization of the resulting <a>normalized dataset</a>, allowing for the isomorphism and digital signing use cases. As blank node identifiers can be stable even with other changes to a graph (dataset), in some cases it is possible to compute the difference between two graphs (datasets), for example if changes are made only to ground triples, or if new blank nodes are introduced which do not create an automorphic confusion with other existing blank nodes. If any information which would change the generated blank node identifier, a resulting diff might indicate a greater set of changes than actually exists.</p>
@@ -338,22 +338,22 @@
   <section id="canon-overview" class="informative">
     <h3>Overview</h3>
 
-    <p>To determine a canonical labeling, the <a>URDNA2015</a> considers the
+    <p>To determine a canonical labeling, <a>URDNA2015</a> considers the
       information connected to each blank node.
-      Nodes with unique first degree information can be issued a canonical identifier
-      immediately via the <a href="#issue-identifier">Issue Identifier algorithm</a>.
+      Nodes with unique first degree information can immediately be issued a canonical identifier
+      via the <a href="#issue-identifier">Issue Identifier algorithm</a>.
       When a node has non-unique first degree information,
-      it is necessary to determine all information that is connected
-      to it transitively throughout the entire dataset.
+      it is necessary to determine all information that is transitively connected
+      to it throughout the entire dataset.
       <a href="#hash-1d-quads" class="sectionRef"></a> defines a
       node’s first degree information via its first degree hash. </p>
 
     <p>Hashes are computed from the information of each blank node.
-      In particular, these hashes encode the mentions incident to each blank node.
+      These hashes encode the mentions incident to each blank node.
       The <a>hash</a> of a string <var>s</var>, is the lower-case,
       hexidecimal representation of the result of passing <var>s</var>
       through a cryptographic hash function.
-      URDNA2015 uses the hash algorithm SHA-256.</p>
+      URDNA2015 uses the SHA-256 hash algorithm.</p>
   </section>
 
   <section id="canon-terms">
@@ -474,13 +474,13 @@
         where the graph component `g` is empty if and only if the
         <a>triple</a> `&lt; s, p, o >` is in the <a>default graph</a>.
         This algorithm considers an RDF dataset to be a set of quads.
-        Two RDF datasets are considered to be the same modulo blank nodes,
-        or isomorphic, if and only if they return the same canonically labeled list of quads
+        Two RDF datasets are considered to be isomorphic (i.e., the same modulo blank nodes),
+        if and only if they return the same canonically labeled list of quads
         via <a>URDNA2015</a>.</p>
 
       <p><a>URDNA2015</a> consists of several sub-algorithms.
         These sub-algorithms are introduced in the following sub-sections.
-        Below we give a high level summary of URDNA2015.</p>
+        First, we give a high level summary of URDNA2015.</p>
 
       <ol>
         <li><strong>Initialization</strong>.
--- a/spec/index.html
+++ b/spec/index.html
@@ -523,16 +523,15 @@
 
       <ol>
         <li id="ca-1">Create the <a>canonicalization state</a>.</li>
-        <li id="ca-2">For every <a>quad</a> in <a>input dataset</a>:
+        <li id="ca-2">For every <a>quad</a> <var>Q</var> in <a>input dataset</a>:
           <ol>
-            <li id="ca-2-1">For each <a>blank node</a> that occurs in the
-              <a>quad</a>, add a reference to the
-              <a>quad</a> using the
+            <li id="ca-2-1">For each <a>blank node</a> that is a component of <var>Q</var>,
+              add a reference to the <var>Q</var> using the
               <a>blank node identifier</a> in the
-              <a>blank node to quads map</a>, creating a new entry if
-              necessary.
+              <a>blank node to quads map</a>,
+              creating a new entry if necessary.
               <div class="issue" data-number="15">
-                It seems that quads must be normalized,
+                It seems that <var>Q</var> must be normalized,
                 so that literals with different syntactic representations
                 but the same semantic representations are merged,
                 and that two graphs differing in the syntactic representation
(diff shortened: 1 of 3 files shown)


gkellogg:
@msporny, @dlongley, I'm pulling from the Arnold/Longley paper, as appropriate. I presume that this is reasonable use, but please let me know if this is okay.

dlongley:
This is looking good to me -- my only concern is with properly crediting @msporny's contributions. We should figure out an appropriate way to do that -- whether that be with "former editors" or something along those lines.

dlongley:
@gkellogg,

> I presume that this is reasonable use, but please let me know if this is okay.

Yes, that's totally fine. Thanks!

gkellogg:
I’m fine with keeping @msporny as a former editor, although that may be intended for former editors of Recs, more formally. I also suggested elsewhere that emerging practice may also list chairs as having a role someplace in the document header.

gkellogg:
@msporny, I used "41758" as your `w3cid`, let me know if this is not correct, or suggest a change.

spec/index.html
```diff
@@ -201,11 +175,38 @@ <h2>Uses of Dataset Canonicalization</h2>
...
+      unless it can be determined that on each occasion,
+      the <a>blank node</a> identifies the same resource;
+      this is due to the fact that <a>blank node identifiers</a>
+      are an aspect of a concrete RDF syntax
+      and are not intended to be persistent or portable.
+      Within the abstract RDF model,
+      blank nodes do not have identifiers
+      (although some
+      <a data-cite="RDF11-CONCEPTS#dfn-rdf-store">RDF store</a>
+      implementations may use stable identifiers and choose to make them portable).
```
TallTed, line 193:
```suggestion
      implementations may use stable identifiers and may choose to make them portable).
```

spec/index.html
```diff
@@ -201,11 +175,38 @@ <h2>Uses of Dataset Canonicalization</h2>
...
+      See <a data-cite="RDF11-CONCEPTS#section-blank-nodes">Blank Nodes</a>
+      in [[!RDF11-CONCEPTS]] for more information.</p>
+
+    <p>RDF does have a provision for allowing blank nodes
+      to be published in an externally identifiable way through the use of
+      <a data-cite="RDF11-CONCEPTS#dfn-skolem-iri">Skolem IRIs</a>,
+      which allows a given RDF store to replace the use of blank nodes
```
TallTed, line 200:
```suggestion
      which allow a given RDF store to replace the use of blank nodes
```

spec/index.html
```diff
@@ -201,11 +175,38 @@ <h2>Uses of Dataset Canonicalization</h2>
...
+      in a concrete syntax with <a>IRIs</a>,
+      which serve to repeatably identify that blank node within that particular RDF store,
```
TallTed, line 202:
```suggestion
      which then serve to repeatably identify that blank node within that particular RDF store;
```

spec/index.html
```diff
@@ -201,11 +175,38 @@ <h2>Uses of Dataset Canonicalization</h2>
...
+      however, this is not generally useful for talking about the
+      same graph in different RDF stores,
+      or other concrete representations.
+      In any case, a stable <a>blank node identifier</a> defined for one
+      RDF store, serialization is arbitrary,
```
TallTed, line 207:
```suggestion
      RDF store or serialization is arbitrary,
```

spec/index.html
```diff
@@ -326,7 +335,28 @@ <h2>Canonicalization</h2>
     This spec should describe the theoretical basis for graph canonicalization and describe
     behavior using normative statements. The explicit algorithms should follow as an informative appendix.</p>
 
-  <section>
+  <section id="canon-overview" class="informative">
+    <h3>Overview</h3>
+
+    <p>To determine a canonical labeling, the <a>URDNA2015</a> considers the
```
TallTed, line 341:
```suggestion
    <p>To determine a canonical labeling, <a>URDNA2015</a> considers the
```

spec/index.html
```diff
@@ -326,7 +335,28 @@ <h2>Canonicalization</h2>
...
+      information connected to each blank node.
+      Nodes with unique first degree information can be issued a canonical identifier
+      immediately via the <a href="#issue-identifier">Issue Identifier algorithm</a>.
```
TallTed, line 344:
```suggestion
      Nodes with unique first degree information can immediately be issued a canonical identifier
      via the <a href="#issue-identifier">Issue Identifier algorithm</a>.
```

spec/index.html
```diff
@@ -326,7 +335,28 @@ <h2>Canonicalization</h2>
...
+      When a node has non-unique first degree information,
+      it is necessary to determine all information that is connected
+      to it transitively throughout the entire dataset.
```
TallTed, line 347:
```suggestion
      it is necessary to determine all information that is transitively connected
      to it throughout the entire dataset.
```

spec/index.html
```diff
@@ -326,7 +335,28 @@ <h2>Canonicalization</h2>
...
+      <a href="#hash-1d-quads" class="sectionRef"></a> defines a
+      node’s first degree information via its first degree hash. </p>
+
+    <p>Hashes are computed from the information of each blank node.
+      In particular, these hashes encode the mentions incident to each blank node.
```
TallTed, line 352:
```suggestion
      These hashes encode the mentions incident to each blank node.
```

spec/index.html
```diff
@@ -326,7 +335,28 @@ <h2>Canonicalization</h2>
...
+      The <a>hash</a> of a string <var>s</var>, is the lower-case,
+      hexidecimal representation of the result of passing <var>s</var>
+      through a cryptographic hash function.
+      URDNA2015 uses the hash algorithm SHA-256.</p>
```
TallTed, line 356:
```suggestion
      URDNA2015 uses the SHA-256 hash algorithm.</p>
```

spec/index.html
```diff
@@ -435,11 +465,47 @@ <h2>Canonicalization Algorithm</h2>
...
       <h3>Overview</h3>
+
+      <p><a>URDNA2015</a> canonically labels an <a>RDF dataset</a>.
+        In URDNA2015, an RDF dataset <var>D</var>
+        is represented as a set of <a>quads</a> of the form `&lt; s, p, o, g >`
+        where the graph component `g` is empty if and only if the
+        <a>triple</a> `&lt; s, p, o >` is in the <a>default graph</a>.
+        This algorithm considers an RDF dataset to be a set of quads.
+        Two RDF datasets are considered to be the same modulo blank nodes,
+        or isomorphic, if and only if they return the same canonically labeled list of quads
```
TallTed, line 478:
```suggestion
        Two RDF datasets are considered to be isomorphic (i.e., the same modulo blank nodes),
        if and only if they return the same canonically labeled list of quads
```

spec/index.html
```diff
@@ -435,11 +465,47 @@ <h2>Canonicalization Algorithm</h2>
...
+        via <a>URDNA2015</a>.</p>
+
+      <p><a>URDNA2015</a> consists of several sub-algorithms.
+        These sub-algorithms are introduced in the following sub-sections.
+        Below we give a high level summary of URDNA2015.</p>
```
TallTed, line 483:
```suggestion
        First, we give a high level summary of URDNA2015.</p>
```

spec/index.html
```diff
@@ -435,11 +465,47 @@ <h2>Canonicalization Algorithm</h2>
...
+
+      <ol>
+        <li><strong>Initialization</strong>.
+          Initialize the state needed for the rest of the algorithm
+          using <a href="#canon-state" class="sectionRef"></a>.</li>
+        <li><strong>Compute first degree hashes</strong>.
+          Compute the first degree hash for each blank node in the dataset using <a href="#hash-1d-quads" class="sectionRef"></a>.</li>
+        <li><strong>Canonically label unique nodes</strong>.
+          Assign canonical identifiers via <a href="#issue-identifier" class="sectionRef"></a>,
+          in lexicographical order, to each blank node whose first degree hash is unique.</li>
```
TallTed, line 493: How is `lexicographical order` defined?  If I understand this doc correctly as it stands, I think including this definition is important, especially because we're working with global data and users, so need to consider the complications inherent in i18n and L10n (i.e., internationalization and localization, or internationalisation and localisation, depending on your locale ... noting also that [W3C uses US English](https://www.w3.org/Guide/manual-of-style/#Spelling)).

((tangent ... it's too bad that [`Append ",spell" to a W3C URI to invoke W3C's spell checker`](https://www.w3.org/Guide/manual-of-style/#Spelling) can't be brought to bear on GitHub-hosted documents.  I wonder whether W3C's tech magicians can provide the dictionary on which that checker must be based for use on GitHub and/or local docs and/or etc?))

TallTed (changes requested):
Mostly minor, except the i18n and L10n concerns

spec/index.html
gkellogg: The [Unicode Collation Algorithm 15.0](https://unicode.org/reports/tr10/) defines a sorting algorithm, which can be pretty complex for encodings other than UTF-8.

Although I've always understood the meaning of "Lexicographical order" as clear (there is a [Wikipedia entry on it](https://en.wikipedia.org/wiki/Lexicographic_order), we can certainly define the term and reference Unicode ordering.

Note that we used the term "lexicographical order" and similar in JSON-LD API with no apparent misunderstanding.

TallTed: > [@gkellogg] 
> The [Unicode Collation Algorithm 15.0](https://unicode.org/reports/tr10/) defines a sorting algorithm, which can be pretty complex for encodings other than UTF-8.

Given that we're already going to require a number of transformations, I think requiring other encodings to be converted to UTF-8 is not unreasonable.  Having skimmed the Unicode sorting algorithm, I'm afraid it will be pretty complex even for UTF-8 encoded data.  (I certainly don't fully grasp it yet!)  Their discussion of [Deterministic Sorting](https://unicode.org/reports/tr10/#Deterministic_Sorting) seems particularly important for implementers of our work.

> [@gkellogg] 
> there is a [Wikipedia entry on [Lexicographical order]](https://en.wikipedia.org/wiki/Lexicographic_order)

Indeed there is ... and its second paragraph starts with —
> [**Wikipedia**] 
> [There are several variants and generalizations of the lexicographical ordering.](https://en.wikipedia.org/wiki/Lexicographic_order#:~:text=There%20are%20several%20variants%20and%20generalizations%20of%20the%20lexicographical%20ordering.)

— which is reinforced by the [several See Also items](https://en.wikipedia.org/wiki/Lexicographic_order#See_also) —

> [**Wikipedia**] 
> * [Collation](https://en.wikipedia.org/wiki/Collation)
> * [Kleene–Brouwer order](https://en.wikipedia.org/wiki/Kleene%E2%80%93Brouwer_order)
> * [Lexicographic preferences](https://en.wikipedia.org/wiki/Lexicographic_preferences)
> * [Lexicographic order topology on the unit square](https://en.wikipedia.org/wiki/Lexicographic_order_topology_on_the_unit_square)
> * [Lexicographic ordering in tensor abstract index notation](https://en.wikipedia.org/wiki/Abstract_index_notation#Braiding)
> * [Lexicographically minimal string rotation](https://en.wikipedia.org/wiki/Lexicographically_minimal_string_rotation)
> * [Long line (topology)](https://en.wikipedia.org/wiki/Long_line_(topology))
> * [Lyndon word](https://en.wikipedia.org/wiki/Lyndon_word)
> * [Star product](https://en.wikipedia.org/wiki/Star_product), a different way of combining partial orders
> * [Orders on the Cartesian product of totally ordered sets](https://en.wikipedia.org/wiki/Total_order#Orders_on_the_Cartesian_product_of_totally_ordered_sets)

> [@gkellogg] 
> Note that we used the term "lexicographical order" and similar in JSON-LD API with no apparent misunderstanding.

This the sort of thing that tends not to get raised as an issue until two people with different interpretations of the meaning of a term trade data and discover their unexpected differences with varying severity of impact.  I regret that I was not involved sufficiently with the JSON-LD API work to have caught this as an issue.

gkellogg:
I've incorporated most of the text I thought was straightforward from Arnold/Longley. Still need examples and images, but I think this is a good point to finish this particular PR.

spec/index.html
```diff
@@ -811,12 +835,128 @@ <h3>Overview</h3>
...
+
+      <p>The following provides a high level outline for how the N-degree hash of <var>n</var>
+        is computed along the shortest <a>gossip path</a>.
+        Note that the full algorithm considers all gossip paths,
+        ultimately returning the <a>hash</a> of the shortest encoded path.</p>
+
+      <ol>
+        <li id="ndh-hl-s1"><strong>Compute related hashes</strong>.
+          Compute the related hash <var>H<sub>n</sub></var> set for <var>n</var>,
+          i.e. all first degree <a>mentions</a> between <var>n</var> and another blank node.
```
TallTed, line 892:
```suggestion
          i.e., all first degree <a>mentions</a> between <var>n</var> and another blank node.
```

spec/index.html
gkellogg: > > [@gkellogg]
> > The [Unicode Collation Algorithm 15.0](https://unicode.org/reports/tr10/) defines a sorting algorithm, which can be pretty complex for encodings other than UTF-8.
> 
> Given that we're already going to require a number of transformations, I think requiring other encodings to be converted to UTF-8 is not unreasonable. Having skimmed the Unicode sorting algorithm, I'm afraid it will be pretty complex even for UTF-8 encoded data. (I certainly don't fully grasp it yet!) Their discussion of [Deterministic Sorting](https://unicode.org/reports/tr10/#Deterministic_Sorting) seems particularly important for implementers of our work.

I expect `String.sort()` to do the right thing on most platforms. It's beyond our scope to exhaustively explore problem areas that belong to the Unicode specs.

See if the definitions I added adequately satisfy your concerns. I really think that we could go overboard trying to specify what lexicographical ordering of Unicode strings means.

gkellogg: Also, note that sorting is done after conversion to N-Quads, which requires quads to be encoded in UTF-8, which simplifies the collation problem.

dlongley: > I expect `String.sort()` as implemented on most platforms to do the right thing on most platforms.

I haven't looked at what the `Unicode Collation Algorithm 15.0` does yet -- but we should definitely make sure that whatever we're specifying matches existing behavior (or is very close and easy to implement w/o performance issues) in languages.

I think, ideally, we'd just be sorting based on Unicode code points. Unfortunately, I think JavaScript compares on code units instead -- and since it uses UTF-16, there may be a need to adjust for cases with surrogate pairs (where the first code unit is `>= 0xD800` or something like that). There's some discussion of sorting adjustments for UTF-16 here:

https://icu-project.org/docs/papers/utf16_code_point_order.html

I'm guessing Java also sorts by default using UTF-16 code units since, IIRC, it represents strings internally that way as well. I'm guessing other languages would sort by UTF-8 code units, but my understanding is that this sort order is the same as sorting by code points.

I'd expect that it would be a mistake to specify that the sort is by UTF-16 code units, so languages that do so natively might need a patch for a certain range. Unless the majority of languages do sorting that way already. Perhaps those languages could internally convert each N-Quad string to UTF-8 byte arrays and then sort by byte for byte -- or take the approaches in that link or something similar. I'm not sure what would be the most efficient solution is, but hopefully we can avoid a significant performance hit.

dlongley: Well, it looks like you can specify `ducet` (Default Unicode Collation Element Table (DUCET)) as the locale to use the Unicode Collation Algorithm in JavaScript ... if it's supported:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/collation

Unfortunately, it also looks like it's perhaps not supported in browsers (well, not in Firefox anyway, if MDN docs are accurate).

dlongley: Also, for our particular use case, it's important that we don't sort based on locale, but rather, ensure that the sort order is the same regardless of locale. This points to using code points or code units to me.

IMO, our requirements are:

1. Consistent sort order no matter the locale.
2. Speed.
3. Simplicity to implement.
4. Matching (or getting a close as possible) to existing implementations.

TallTed:
> we could go overboard trying to specify what lexicographical ordering of Unicode strings means

Either the ordering is important and necessary, or it isn't.  If it isn't, then all reference to it should be removed.  If it is, either we need to provide sufficient detail that the ordering is predictable and consistent, or we need to explain why (and how much of) the unpredictability and/or inconsistency is acceptable.

Presuming the ordering is important and necessary, I think @dlongley's suggestion of UTF-8 codepoint order is likely to play an important part. (It was in my own thinking yesterday, but not firm enough to write up, given the other questions still in the air.)

I'm pretty sure we'll need more discussion on this, but it probably does need its own issue(s).  For now, I'm happy enough that it's on more people's radar than my own.

gkellogg:
Ensuring proper ordering, whatever it may be, would be important regardless of what algorithm is chosen for actually navigating the quads to create hashes. I hadn't looked at the documents closely enough to be sensitive to the Localization issues, and wasn't aware of the JavaScript implementation issues @dlongley pointed out.

Given the fact that this is something which would face many different specifications that there is not a single solution useable across domains. This may be a case where we call for early feedback from either the Internationalization WG or the TAG.

I'll create an issue to track.

spec/index.html
afs: Unicode collation is not helpful. It is locale sensitive and you must know the locale. It is not total.

> https://unicode.org/reports/tr10/#Common_Misperceptions
> 
> "Swedish and German share most of the same characters, for example, but have very different sorting orders."
> "Collation is not a property of strings."
> "Collation order is not preserved when comparing sort keys generated from different collation sequences."

For @dlongley points 1-4, the conclusion is one of codepoint sorting (logically convert to 21 bits values, sort) or 
a code unit for a specific code unit choice.

SPARQL is (strictly) codepoint (the abstract character) ordering.

[Section 17.3](https://www.w3.org/TR/sparql11-query/#OperatorMapping) says use collation `http://www.w3.org/2005/xpath-functions/collation/codepoint` (credit to Eric Prud'hommeaux).

dlongley: I also suspect we want to compare using code points. The only reason we'd choose code units would be if most languages already natively sort by UTF-16 code units, but I think this is unlikely.

For those languages that don't offer some fast, native support for code point sorting, but instead compare natively using UTF-16 code units, I imagine the input could be scanned just once to look for surrogate pairs. Only if any were detected would special comparison code need to be used.

This discussion is somewhat split between here and #18 -- I recommend we say in this PR that lexicographical order refers to Unicode code point order and then continue the discussion over there. My guess is that existing implementations already sort by Unicode code point order or are *very close* and just need a slight adjustment for some data because they compare using UTF-16 code units today.

afs: Does DUCET canonicalize codepoints? (https://unicode.org/reports/tr10/#Contractions_DUCET)

So different ways of writing the same character (e.g. combining characters, [Unicode Normalization](https://unicode.org/reports/tr15/)) have the same weight but then two different xsd:strings are not totally ordered if one contains the character as one codepoint and the other used two (the second being a combining character).

Not simple any more!

msporny (approved):

spec/index.html
```diff
@@ -369,12 +369,18 @@ <h2>Canonicalization Algorithm Terms</h2>
...
       <dt><dfn>hash algorithm</dfn></dt>
       <dd>The hash algorithm used by <a>URDNA2015</a>, namely, SHA-256.</dd>
-      <dt><dfn>lexicographical order</dfn></dt>
-      <dd>This refers to determining the order of two Unicode strings,
-        as defined in [[UTS10]].
+      <dt><dfn data-lt="lexicographical|lexicographically">lexicographical order</dfn></dt>
+      <dd>This refers to determining the order of two Unicode strings (`A` and `B`),
+        by using <br/><code><a data-cite="XPATH-FUNCTIONS#func-compare">fn:compare</a>(A, B, "http://www.w3.org/2005/xpath-functions/collation/codepoint")</code>,
+        as defined in [[XPATH-FUNCTIONS]].
+        This defines an absolute ordering of Unicode strings independent of language and localle.
```
TallTed, line 376:
```suggestion
        This defines an absolute ordering of Unicode strings independent of language and locale.
```

spec/index.html
```diff
@@ -1015,7 +1021,7 @@ <h3>Algorithm</h3>
         </li>
         <li id="hndq-4">Create an empty string, <var>data to hash</var>.</li>
         <li id="hndq-5">For each <var>related hash</var> to <var>blank node list</var> mapping in
-          <var>hash to related blank nodes map</var>, sorted lexicographically
+          <var>hash to related blank nodes map</var>, <a>lexicographically-sorted</a>
```
TallTed, line 1024:
```suggestion
          <var>hash to related blank nodes map</var>, <a>lexicographically sorted</a>
```

TallTed (changes requested):
A couple of small tweaks.

dlongley (approved):

spec/index.html
gkellogg: I did this comprehensively.

gkellogg:
@msporny I added the `issue-summary` section, but it looks like it just picked up the issues explicitly mentioned in the spec. Couldn't see anything about configuring to pull all issues from the repo.

spec/index.html
```diff
@@ -712,7 +712,7 @@ <h3>Overview</h3>
         Every blank node component is replaced with either `_:a` or `_:z`,
         depending on if that component is <var>n</var> or not.</p>
 
-      <p>The resulting serialized quads are then <a>lexicographically-sorted</a>,
+      <p>The resulting serialized quads are then <a>lexicographically sorted</a>,
         concatenated and hashed.
```
TallTed, line 716:
```suggestion
        concatenated, and hashed.
```

afs (approved):
"Approved" but I think this is a good time to start to remove the terminology "lexicographic order".  It is confusing because it may be taken to imply it works on characters, not codepoints (combining characters etc), and hence unicode normalization and possibly locale sensitive ordering is being specified when it's not.

I think we settled on needing a fixed [total ordering](https://en.wikipedia.org/wiki/Total_order) for any strings (URIs or literals) by using codepoint ordering which has no semantic (string value, or collation) semantics.

For speed, at least an issue box saying we will remove the use of "lexicographic order" and replace it with "[total order](https://en.wikipedia.org/wiki/Total_order) of RDF terms".

gkellogg:
I can do a commit changing "lexicographical order" (and related) with "total order" (or "codepoint order") later this weekend.

@dlongley WDYT?

dlongley:
@gkellogg,

Yes, I think it's a good idea to change the language. My vote is for "Unicode code point order" for clarity.

afs:
No need to define it -- "[total ordering](https://en.wikipedia.org/wiki/Total_order)" (wikipedia link) is the term for the requirement we need (and not some weird lattice comparison).

"Unicode code point order" seems the best choice for a web standard.

spec/index.html
```diff
@@ -1074,14 +1071,14 @@ <h3>Algorithm</h3>
...
-                      <var>permutation</var>.</li>
+                      of <var>chosen path</var> and <var>path</var> is
+                      greater than <var>chosen path</var> when consider <a>code point order</a>,
+                      then skip to the next <var>permutation</var>.</li>
                   </ol>
                 </li>
                 <li id="hndq-5-4-6">If <var>chosen path</var> is empty or <var>path</var> is
-                  <a>lexicographically</a> less than <var>chosen path</var>, set
-                  <var>chosen path</var> to <var>path</var> and <var>chosen issuer</var>
+                  less than <var>chosen path</var>  when consider <a>code point order</a>,
```
dlongley, line 1080:
```suggestion
                  less than <var>chosen path</var> when considering <a>code point order</a>,
```

spec/index.html
```diff
@@ -883,7 +879,7 @@ <h3>Overview</h3>
         <li>The length of <var>s</var> is less than or equal to the length
```
dlongley, line 879: There's a typo "shortest1" a couple lines above this one.

spec/index.html
```diff
@@ -883,7 +879,7 @@ <h3>Overview</h3>
...
           of any other gossip path string <var>s′</var>.</li>
         <li>If <var>s</var> and <var>s′</var> have the same length (as strings),
-          then <var>s</var> is <a>lexicographically</a> less than or equal to <var>s′</var>.</li>
+          then <var>s</var> <a>code point orders</a> less than or equal to <var>s′</var>.</li>
```
dlongley, line 882: This reads a little funny to me ... should it be this?:

```suggestion
          then <var>s</var> is <a>code point ordered</a> less than or equal to <var>s′</var>.</li>
```

spec/index.html
```diff
@@ -1074,14 +1071,14 @@ <h3>Algorithm</h3>
                       <a>identifier issuer</a> in result.</li>
                     <li id="hndq-5-4-5-5">If <var>chosen path</var> is not empty and the length
                       of <var>path</var> is greater than or equal to the length
-                      of <var>chosen path</var> and <var>path</var> is <a>lexicographically</a>
-                      greater than <var>chosen path</var>, then skip to the next
-                      <var>permutation</var>.</li>
+                      of <var>chosen path</var> and <var>path</var> is
+                      greater than <var>chosen path</var> when consider <a>code point order</a>,
```
dlongley, line 1075:
```suggestion
                      greater than <var>chosen path</var> when considering <a>code point order</a>,
```

Alternatively, we could ```<var>path</path> is <a>code point ordered</a> greater than``` ... right?

dlongley (approved):
Approving with some suggestions to consider. Thanks!

PR #22, opened 2022-11-01
Referencing issues

philarcher:
References to specific issues added in relevant locations in the doc (especially those related to algorithm choices) in prep for FPWD

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -464,6 +464,14 @@
 
   <section id="canon-algorithm" class="algorithm">
     <h2>Canonicalization Algorithm</h2>
+  <p class="ednote">At the time of writing, there are several open issues that will determine important details of the canonicalization algorithm.</p>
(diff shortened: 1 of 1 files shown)


philarcher:
I'm not concerned whether this specific PR is merged so much as ensuring that the issues are cited in the doc in the relevant places, especially around the actual algorithm. That's where there are several open issues that will determine much of the future direction of the spec. The basic message to the world is "we have a (very) good starting point but we're not done yet - these are the kind of things we're asking ourselves". WDYT?

gkellogg (approved):
I had thought that the issue-summary section would automatically bring in open skies, but it doesn’t seem to do so. It is something of a pain to manually manage this list, but it works.

I’ve thought of adding a Design Considerations section where we could reference resolved issues with an abbreviated decision summary.

PR #26, opened 2022-11-04
Add two examples for Hash First Degree Quads,

gkellogg:
illustrating the behavior of the algorithm.

I added a diagrams for the examples.

A more automated way of invoking the algorithm to calculate/validate the results would be useful. ~~But, note that the example quads aren't actually valid, and the results are calculated over these string representations, rather than applying a base IRI and legitimate N-Quads serialization, but they do illustrate the operation of the algorithm.~~

The description for the figures may need more expansion and a descriptive appending to satisfy accessibility requirements of the spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rch-rdc/pull/26.html" title="Last updated on Nov 7, 2022, 9:12 PM UTC (c30a029)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rch-rdc/26/0e7326d...c30a029.html" title="Last updated on Nov 7, 2022, 9:12 PM UTC (c30a029)">Diff</a>


gkellogg:
Although this was not intended for FPWD, I think it generally improves both the algorithm descriptions and provides some useful examples. We might consider it for FPWD, but it's not a critical addition, IMHO.

gkellogg:
Note, the diagrams don't come through in the Preview, but you can find them [here](https://docs.google.com/drawings/d/1LQ6fp4a35lrEOKRMD20gcncUkoitX0eW8QI_U2mQnAs) and [here](https://docs.google.com/drawings/d/1HWpz8S-PNaWi02-utF62HAMd_0b-AF_tHNSRRUQILbU), largely based on one from Aiden.

spec/index.html
```diff
@@ -734,9 +735,153 @@ <h3>Overview</h3>
...
+
+        <pre data-transform="updateExample">
+          <!--
+            :p :q _:a .
+            _:a :p _:z .
+          -->
+        </pre>
+
+        <p>These are then serialized to N-Quads, concatenated and hashed resulting in
+          `3b26142829b8887d011d779079a243bd61ab53c3990d550320a17b59ade6ba36`.</p>
```
dlongley, line 784: If this is a real hash, we should probably say that the hash algorithm in use here is SHA-256. At a later time we should do a pass over the spec to make sure we can swap the hash algorithm out for something else in the future (which would create a "new canonicalization algorithm", but only different in the hash algorithm chosen).

spec/index.html
```diff
@@ -734,9 +735,153 @@ <h3>Overview</h3>
...
+
+        <pre data-transform="updateExample">
+          <!--
+            :p :r _:a .
+            _:a :p _:z .
+          -->
+        </pre>
+
+        <p>These are then serialized to N-Quads, concatenated and hashed resulting in
+          `3b26142829b8887d011d779079a243bd61ab53c3990d550320a17b59ade6ba36`.</p>
```
dlongley, line 797: This hash is the same as the above; even if it's just an example here it should be different because the example is about  the hash being different.

spec/index.html
```diff
@@ -734,9 +735,153 @@ <h3>Overview</h3>
...
+             title="Graph with blank nodes resulting in unique hashes">
+          <!--
+            :p :q _:b0 .
+            :p :r _:b1 .
+            _:b0 :p _:b2 .
+            _:b1 :p _:b2 .
+          -->
+        </pre>
+
+        <p>The algorithm will be called twice, with each blank node (`_:b0` and `_:b1`).
```
dlongley, line 772: There's a third blank node here `_:b2`. Should that instead be a non-blank node in this example?

spec/index.html
```diff
@@ -734,9 +735,153 @@ <h3>Overview</h3>
...
+          <!--
+            :p :r _:a .
+            _:a :p _:z .
+          -->
+        </pre>
+
+        <p>These are then serialized to N-Quads, concatenated and hashed resulting in
+          `3b26142829b8887d011d779079a243bd61ab53c3990d550320a17b59ade6ba36`.</p>
+
+        <p>Thus the generated hashes each reference just a single blank node,
```
dlongley, line 799: See above -- the hashes reference multiple blank nodes -- but only because of `_:b2` which could become a non-blank node to remedy that.

spec/index.html
gkellogg: Would a reference to "hash algorithm", which does say that it is SHA-256 work for you? Or, would you rather replace the actual hash value with something more abstract? My thought was that we should strive to make the examples accurate, so that we could potentially automatically check for correctness.

spec/index.html
gkellogg: This was due to an error in my manual creation of the hashes. The following should e correct.

```suggestion
          `15973d39de079913dac841ac4fa8c4781c0febfba5e83e5c6e250869587f8659`.</p>
```

spec/index.html
gkellogg: You're correct; the example evolved from the original I had. But `_:b0 :p _:b2` and `_:b1 :p _:b2` would exhibit the more complicated scenario of requiring an N-Degree hash, which I was introducing in the next example. Better to replace the graph with something not requiring an N-Degree hash.

spec/index.html
gkellogg: You're correct; the example evolved from the original I had. But `_:b0 :p _:b2` and `_:b1 :p _:b2` would exhibit the more complicated scenario of requiring an N-Degree hash, which I was introducing in the next example. Better to replace the graph with something not requiring an N-Degree hash.

spec/index.html
gkellogg: Following your suggestion, I'll re-work this example to the following:

```
:p :q _:b0 .
:p :r _:b1 .
_:b0 :p :s .
_:b1 :r :s .
```

This should meet the original intention of this example, to have two blank nodes which can be uniquely hashed and not require the use of the N-Degree hash algorithm.

spec/index.html
```diff
@@ -734,9 +735,153 @@ <h3>Overview</h3>
...
+        <p>Using the reference node `_:b2` results in the following quads:</p>
+
+        <pre data-transform="updateExample">
+          <!--
+            _:z :p _:a .
+            _:a :r _:z .
+          -->
+        </pre>
+
+        <p>Which hashes to: `3b26142829b8887d011d779079a243bd61ab53c3990d550320a17b59ade6ba36`.</p>
```
dlongley, line 865: Some of these hash values that aren't supposed to be the same are shown as the same, so they probably just need recalculation, etc.

spec/index.html
```diff
@@ -977,8 +1130,93 @@ <h3>Overview</h3>
...
+            :p :q ****_:b0**** .
+            :p :q ****_:b1**** .
+            ****_:b0**** :p _:b2 .
+            ****_:b1**** :p _:b3 .
+            _:b2 :r _:b3 .
+          -->
+        </pre>
+
+        <p>The algorithm will be called twice, with each blank node
+          (`_:b0`, `_:b1`) which did not result in a unique hash
```
dlongley, line 1174: I think the algorithm will actually be called 4 times (another time each for `_:b2` and `_:b3`) here. I understand that we're looking to continue the example from earlier so maybe this is meant to be further edited in some way -- just leaving a note here.

dlongley (commented):
Looking better! I left a few more comments -- I think we still need a couple tweaks.

spec/index.html
dlongley: Yeah, a reference would be fine.

spec/index.html
gkellogg: This example should have been left out entirely, but was added back accidentally. Examples for Hash N-Degree Quads need more work, which I was going to do in a different PR later.

dlongley (approved):
Thanks!

PR #27, opened 2022-11-07
Minor fix to issue-identifier-algorithm.

gkellogg:
Wrong variable name used.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rch-rdc/pull/27.html" title="Last updated on Nov 7, 2022, 10:24 PM UTC (7ef6ccc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rch-rdc/27/0e7326d...7ef6ccc.html" title="Last updated on Nov 7, 2022, 10:24 PM UTC (7ef6ccc)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -671,7 +671,7 @@
 
       <ol>
         <li id="iia-1">If there is already an issued identifier for
-          <var>existing identifier</var> in <a>issued identifiers list</a> for <var>existing identifier</var>,
+          <var>existing identifier</var> in <a>issued identifiers list</a> of <var>I</var>,
           return it.</li>
         <li id="iia-2">Generate <var>issued identifier</var> by concatenating
           <a>identifier prefix</a> with the string value of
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -671,7 +671,7 @@ <h3>Algorithm</h3>
 
       <ol>
         <li id="iia-1">If there is already an issued identifier for
-          <var>existing identifier</var> in <a>issued identifiers list</a> for <var>I</var>,
+          <var>existing identifier</var> in <a>issued identifiers list</a> for <var>existing identifier</var>,
```
dlongley, line 674: Hmm, should it be this instead?

```suggestion
          <var>existing identifier</var> in <a>issued identifiers list</a> of <var>I</var>,
```

It seems like the variable wasn't wrong, but the language was. I think what we're trying to check here is if there is an issued identifier for a `existing identifier` in `I's list of issued identifiers`. So we could use `of` or the possessive language I just used -- whichever seems better to you (unless I'm just confused of course!).

dlongley (changes requested):

PR #28, opened 2022-11-09
Short name and alternatives

gkellogg:
This updates references to "rch-rdc" and "rdf-dataset-canonicalization" and adds @yamdan as an editor. It also changes a paragraph in SOTD to reference the issues related to the future selection of an algorithm (word-smithing appreciated).

Note, @yamdan, the editor entry could use information about you including, a personal URL, affiliation, affiliation URL, and most importantly, your w3cid, which I believe is required for publication. @pchampin can probably help in finding this.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/28.html" title="Last updated on Nov 10, 2022, 6:02 PM UTC (573e9e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/28/50485e3...573e9e2.html" title="Last updated on Nov 10, 2022, 6:02 PM UTC (573e9e2)">Diff</a>

PR #30, opened 2022-11-11
Use `https` for test vocabulary

yamdan:
Unless I misunderstood https://github.com/w3c/rdf-canon/pull/28#issuecomment-1310506672, all the places where our test vocabulary (`rdfn:`) appears as `http` URIs should be replaced with `https` URIs for consistency.


davidlehn:
While breaking changes are happening, I believe `rdfn` came from "RDF Graph Normalization", so now it should probably be `rdfc`?

gkellogg:
We can change them, perhaps “rdfc” would be a good prefix. Note that, other than for vocab.til, the changes should really go in the mk_ files, and then regenerate the rest.

dlongley (approved):
LGTM ... `rdfc` as a prefix is fine with me too.

gkellogg:
@yamdan, when you have full access to the repo (nudge-nudge @pchampin) if you do your PR branches on this repo, it allows other editors to update it directly. Also, making a change in your main branch can lead to problems when you want to re-synch with this repo, as it may have diverged from your own.

I think the best thing to do is to re-create the work, with the change to "rdfc" on another branch and substitute that for this PR, rather than merging this and then making more changes in a different PR.

gkellogg:
Closing in favor of #32, thanks @yamdan.

yamdan:
Thank you @gkellogg for your advice and re-creating the PR. Now that I understand the better way to do PR using branches, I won't make the same mistake next time...

PR #31, opened 2022-11-11
Manually add "shrink-to-fit=no" to viewport meta element

gkellogg:
This should be done by ReSpec, but it seems to be missing from the generated output.

@pchampin @iherman I don't know if you've encountered this before, as the only code I see in ReSpec includes this attribute, yet it was missing from the generated FPWD/Overview.html. It could come back to bite is with Echidna in the future.


pchampin:
I believe I found the reason of this problem:
the HTML source does contain a `<meta name=viewport ...>` element, so ReSpec does *not* add its own.
If you remove that element from the source, ReSpec will generate it with the correct value.

gkellogg:
Great, thanks. I’ll update this. I had created a respec issue, too.

PR #35, opened 2022-11-22
Fix broken links

pchampin:
Those broken links were detected by the [Link Checker](https://validator.w3.org/checklink).

@gkellogg  you might probably want to merge this before generating a new FPWD.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/35.html" title="Last updated on Nov 22, 2022, 1:41 PM UTC (afea8e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/35/14c6a48...afea8e9.html" title="Last updated on Nov 22, 2022, 1:41 PM UTC (afea8e9)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -66,7 +66,7 @@
     doJsonLd:             true,
 
     // name (with the @w3c.org) of the public mailing to which comments are due
-    wgPublicList: "public-rch-wg@w3.org"
+    wgPublicList: "public-rch-wg"
   };
 //]]>
 </script>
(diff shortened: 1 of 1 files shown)


gkellogg:
Thanks, I didn’t see these when I ran the link checker.

spec/index.html
```diff
@@ -66,7 +66,7 @@
     doJsonLd:             true,
 
     // name (with the @w3c.org) of the public mailing to which comments are due
-    wgPublicList: "public-rch-wg@w3.org"
+    wgPublicList: "public-rch-wg"
```
TallTed, line 69: Not that it matters much at the moment, but I'm thoroughly confused by these lines.

The comment-instruction *appears* to say that `public-rch-wg@w3.org` is the desired value (being `with the @w3c.org`), but the change just merged is to `public-rch-wg`.

If this is the intended meaning of the instruction, I suggest logging a PR against their source, to change `with` to `minus`, and optimally to add an example.

gkellogg: I'm not sure the source of that original comment, but the [wiki entry](https://github.com/w3c/respec/wiki/wgPublicList) doesn't use the `@w3.org`. I'll remove the comment.

PR #40, opened 2022-11-23
Removed the simple loop - issue 23

### LINKED ISSUE #23: Explore and resolve `simple` loop editor's notes
There are two related editor's notes in the URDN2015 spec about looping over the first step of the algorithm. The first note is in the "4.4 Canonicalization Algorithm" section and the second in the "4.6 Hash First Degree Quads" section.

These notes are about a mistake that was made during the initial implementation and roll out of the URDNA2015 spec many years ago. The first step of the algorithm that leverages the "Hash First Degree Quads" was meant to be looped over multiple times while canonical labels could be assigned based on "simple" differences. Once no more "simple" differences could be determined, that loop would end and the second step of the algorithm would be run.

However, the "Hash First Degree Quads" function in the spec did not include text to use these canonical bnode labels as they were assigned, when generating hashes. Implementations did use them at some point, but in an effort to conform to the spec (or some other refactoring probably lost to history) removed them -- and the algorithm instead never considers canonical bnode labels. This means that the hash output never changes on account of them as they are assigned. Therefore, the "simple" loop in section 4.4 will only ever run once. Therefore, the algorithm text could remove the loop without affecting the algorithm outputs.

The end result of this is that some differences in datasets that are simple enough to detect without moving into the second step of the algorithm (which recursively and comprehensively considers all paths / relationships) aren't caught early. They, of course, are still handled by the second step of the overall algorithm, but originally they were meant to be handled by looping over the simpler step.

The URDNA2015 spec text today is still "correct" (in that it matches the test suite and implementations in the wild), it could just be simplified because the "simple loop" isn't actually used. Alternatively, this missing optimization could be added to produce an algorithm (with different outputs), but that would come at some cost to existing implementations and signed documents out in the wild. It's not clear how costly it would be -- there may not be that much data that ever needed a second run through the "simple loop", just like there is very little data that is expected to need the second step at all.

iherman:
The core algorithm is now without the extra loop controlled by `simple`.

This is to fix #23 (if there is a consensus about this, of course).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/40.html" title="Last updated on Nov 23, 2022, 7:04 PM UTC (0dc13ed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/40/22e6c14...0dc13ed.html" title="Last updated on Nov 23, 2022, 7:04 PM UTC (0dc13ed)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1320,20 +1320,20 @@
 <section class="appendix informative" id="changes-from-fpwd">
   <h2>Changes since the First Public Working Draft of 24 November 2022</h2>
   <ul>
-    <li class="issue" data-number="23">
+    <li>
       <a href="#hash-1d-quads" class="sectionRef"></a>
       was simplified to remove the `simple` flag,
       which was unused in existing implementations.
-      The design of the algorithm was to use the
+      The original design of the algorithm was to use the
       assigned canonical <a>blank node identifier</a>,
       if available, instead of `_:a` or `_:z`,
       similar to how it is used in
       the related hash algorithm, but this text never made it into the spec
       before implementations moved forward.
-      Therefore, the hashes here
-      never change, making the loop based on the `simple`
-      flag that calls this algorithm unnecessary; it needs to only run
-      once.
+      Therefore, the hashes never change,
+      making the loop based on the `simple`
+      flag that calls this algorithm unnecessary.
+      See <a href="https://github.com/w3c/rdf-canon/issues/23">Issue 23</a> for the discussion.
     </li>
   </ul>
 </section>
--- a/spec/index.html
+++ b/spec/index.html
@@ -576,7 +576,6 @@
               <a>hash to blank nodes map</a>.</li>
           </ol>
         </li>
-        </li>
         <li id="ca-6">For each <var>hash</var> to <var>identifier list</var>
           <a data-cite="INFRA#map-entry">map entry</a> in
           <a>hash to blank nodes map</a>, <a>code point ordered</a> by
@@ -1337,7 +1336,6 @@
     </li>
   </ul>
 </section>
-<section id="issue-summary" class="appendix informative"></section>
 
 <section id="ack" class="appendix informative">
   <h2>Acknowledgements</h2>
--- a/spec/index.html
+++ b/spec/index.html
@@ -542,76 +542,66 @@
         <li id="ca-3">Create a list of non-normalized <a>blank node identifiers</a>
           <var>non-normalized identifiers</var> and populate it using the keys
           from the <a>blank node to quads map</a>.</li>
-        <li id="ca-4">Initialize <var>simple</var>, a boolean flag, to
-          <code>true</code>.</li>
-        <li id="ca-5">While <var>simple</var> is <code>true</code>,
-          issue canonical identifiers for <a>blank nodes</a>:
+        <li id="ca-4">Clear <a>hash to blank nodes map</a>.</li>
+        <li id="ca-5">For each <a>blank node identifier</a> <var>n</var> in <var>non-normalized identifiers</var>:
           <ol>
-            <li id="ca-5-1">Set <var>simple</var> to <code>false</code>.</li>
-            <li id="ca-5-2">Clear <a>hash to blank nodes map</a>.</li>
-            <li id="ca-5-3">For each <a>blank node identifier</a>
-              <var>n</var> in <var>non-normalized identifiers</var>:
-              <ol>
-                <li id="ca-5-3-1">Create a <a>hash</a>, <var>h<sub>f</sub>(n)</var>,
-                  for <var>n</var> according to the
-                  <a href="#hash-1d-quads">Hash First Degree Quads algorithm</a>.</li>
-                <li id="ca-5-3-2">Add <var>h<sub>f</sub>(n)</var> and <var>n</var> to
-                  <a>hash to blank nodes map</a>, including repetitions,
-                  creating a new entry if necessary.</li>
-              </ol>
-            </li>
-            <li id="ca-5-4">For each <var>hash</var> to <var>identifier list</var> mapping in
-              <a>hash to blank nodes map</a>, <a>code point ordered</a> by
-              <var>hash</var>:
-              <ol>
-                <li id="ca-5-4-1">If  <var>identifier list</var> has more than one entry,
-                  continue to the next mapping.</li>
-                <li id="ca-5-4-2">Use the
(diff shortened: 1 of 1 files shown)


iherman:
I am not sure how the possible core references are done. I have renumbered the `id` values in the algorithmic steps to be consistent, but I did not check whether there are cross references to those... :-( If those cross references already exist, ping me, and I will go through a check (it just takes time...)

gkellogg:
> I am not sure how the possible core references are done. I have renumbered the `id` values in the algorithmic steps to be consistent, but I did not check whether there are cross references to those... :-( If those cross references already exist, ping me, and I will go through a check (it just takes time...)

Algorithm step identifiers are maintained manually, and benefit from algorithm stability. Past experience has indicated that having such references in place is useful in the long-run, but can be challenging to maintain during flux. I'll give it a once-over.

There are some other references to "simple" and the relating issue that can come out as well, which I'll look at.

spec/index.html
```diff
@@ -543,76 +543,66 @@ <h3>Algorithm</h3>
...
+              <a href="#hash-1d-quads">Hash First Degree Quads algorithm</a>.</li>
+            <li id="ca-5-2">Add <var>h<sub>f</sub>(n)</var> and <var>n</var> to
+              <a>hash to blank nodes map</a>, including repetitions,
+              creating a new entry if necessary.</li>
           </ol>
         </li>
         <li id="ca-6">For each <var>hash</var> to <var>identifier list</var> mapping in
+          <a>hash to blank nodes map</a>, <a>code point ordered</a> by <var>hash</var>:
+          <ol>
+            <li id="ca-6-1">If  <var>identifier list</var> has more than one entry,
```
dlongley, line 560:
```suggestion
            <li id="ca-6-1">If <var>identifier list</var> has more than one entry,
```

dlongley (approved):
I see your comment, @gkellogg, that there's some other simple stuff to be removed -- I'm not sure if you meant in a subsequent PR or this one, but I'll approve this because it looks right for removing the loop.

pchampin:
Mentioned during today's call: https://www.w3.org/2022/11/23-rch-minutes.html#t03

spec/index.html
```diff
@@ -543,76 +543,66 @@ <h3>Algorithm</h3>
         <li id="ca-3">Create a list of non-normalized <a>blank node identifiers</a>
           <var>non-normalized identifiers</var> and populate it using the keys
           from the <a>blank node to quads map</a>.</li>
-        <li id="ca-4">Initialize <var>simple</var>, a boolean flag, to
-          <code>true</code>.</li>
-        <li id="ca-5">While <var>simple</var> is <code>true</code>,
-          issue canonical identifiers for <a>blank nodes</a>:
+        <li id="ca-4">Clear <a>hash to blank nodes map</a>.</li>
```
yamdan, line 546: Now we are not in the while loop, it might be better to use "Create" just like the other parts in this document.
```suggestion
        <li id="ca-4">Create <a>hash to blank nodes map</a>.</li>
```

yamdan (approved):

gkellogg:
> I see your comment, @gkellogg, that there's some other simple stuff to be removed -- I'm not sure if you meant in a subsequent PR or this one, but I'll approve this because it looks right for removing the loop.

Just references to the issue, which we agreed to gather in a "Changes Since FPWD" section.

spec/index.html
gkellogg: The **hash to blank nodes map** is part of the canonicalization state, so _creating_ it is probably not the right term to use here, either.

Given that step 1 initializes the canonicalization state, which implicitly describes the hash to blank nodes map, I think that step 4 becomes unnecessary, but can be described in a summary description on step 1 for clarity.

gkellogg (approved):

spec/index.html
gkellogg: @yamdan Please see if the updated algorithm description suites.

spec/index.html
```diff
@@ -393,11 +398,11 @@ <h2>Canonicalization State</h2>
 
     <dl>
       <dt><dfn>blank node to quads map</dfn></dt>
-      <dd>A data structure that maps a <a>blank node identifier</a> to
+      <dd>A <a data-cite="INFRA#ordered-map">map</a> that relates a <a>blank node identifier</a> to
```
dlongley, line 401: I think it would be better to say "ordered map" in the text here, even if the link is right:

```suggestion
      <dd>An <a data-cite="INFRA#ordered-map">ordered map</a> that relates a <a>blank node identifier</a> to
```

spec/index.html
```diff
@@ -393,11 +398,11 @@ <h2>Canonicalization State</h2>
...
         the <a>quads</a> in which they appear in the
         <a>input dataset</a>.</dd>
       <dt><dfn>hash to blank nodes map</dfn></dt>
-      <dd>A data structure that maps a <a>hash</a> to a list of
+      <dd>A <a data-cite="INFRA#ordered-map">map</a> that relates a <a>hash</a> to a list of
```
dlongley, line 405:
```suggestion
      <dd>An <a data-cite="INFRA#ordered-map">ordered map</a> that relates a <a>hash</a> to a list of
```

dlongley (approved):
Thanks! See suggestion for clarity where ordered maps are used in the canonicalization state.

spec/index.html
gkellogg: I don’t think it was ordered before, is order significant here as well? Of course, the citation is to an ordered map in any case.

dlongley: Oh, if it wasn't ordered before, nevermind. I may have mixed this up with the identifier issuer map.

spec/index.html
yamdan: Thank you @gkellogg, I agree with your updated description; I had overlooked that the canonicalization state in step 1 includes the hash to blank nodes map.

spec/index.html
```diff
@@ -522,13 +519,22 @@ <h3>Overview</h3>
       <h3>Algorithm</h3>
 
       <ol>
-        <li id="ca-1">Create the <a>canonicalization state</a>.</li>
+        <li id="ca-1">Create the <a>canonicalization state</a>.
+          <details>
+            <summary>Explaination</summary>
```
yamdan, line 524:
```suggestion
            <summary>Explanation</summary>
```

yamdan (approved):
Thanks a lot. See a suggestion for a small typo.

PR #49, opened 2022-12-04
setup Echidna auto-publishing

pchampin:
since this is the first time I setup echidna, I''d rather have others check what I did.

NB: I have also generated a token and added it as a repository secret.

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -1,0 +1,21 @@
+# Example: Override respecConfig for W3C deployment and validators.
+name: Echidna Auto-publish
+on:
+  pull_request: {}
+  push:
+    branches: [main]
+jobs:
+  main:
+    name: Echidna Auto-publish WD
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: w3c/spec-prod@v2
+        with:
+          TOOLCHAIN: respec
+          SOURCE: spec/index.html
+          W3C_ECHIDNA_TOKEN: ${{ secrets.ECHIDNA_TOKEN }}
+          W3C_WG_DECISION_URL: https://www.w3.org/2022/11/09-rch-minutes#r03
+          W3C_BUILD_OVERRIDE: |
+            specStatus: WD
+            shortName: rdf-canon


pchampin:
I didn't realize that this workflow would also run an PRs, but as it is obviously the case, it seems to be redundant with the `ci.yaml` that was already there... We should probably clean this up eventually.

.github/workflows/auto-publish.yml
```diff
@@ -0,0 +1,21 @@
+# Example: Override respecConfig for W3C deployment and validators.
+name: Echidna Auto-publish
+on:
+  pull_request: {}
+  push:
+    branches: [main]
```
iherman, line 6: This will regenerate the publication on _any_ change in the repository (e.g., any change in the test directory). We may not want to do this. By changing these lines, we may avoid this:
```suggestion
on:
  pull_request:
    paths: ["spec/**"]
  push:
    branches: [main]
    paths: ["spec/**"]
```

@msporny may also want to review this to check if I am right; I am not a very good gh action expert.

iherman (approved):

msporny (approved):
LGTM.

@gkellogg I raised an eyebrow at the `spec/**` structure, and noted that going to the ED doesn't take you to the spec, but a summary page of sorts (rendering of the README.md). I don't know if that was intentional (I expect it was), surprised me a bit, and took me a bit (15 whole seconds of reading, the horror!) to find the link to the spec. 

I'd prefer to link straight to the spec and have links to test suite and other things from the spec. My concern is hitting a page that doesn't look "official" when going to the ED. That said, I don't feel strongly about this, so happy to defer to the Editors and Chairs (and/or WG) on the best set up. Each WG has its own culture. :)

gkellogg:
> LGTM.
> 
> @gkellogg I raised an eyebrow at the `spec/**` structure, and noted that going to the ED doesn't take you to the spec, but a summary page of sorts (rendering of the README.md). I don't know if that was intentional (I expect it was), surprised me a bit, and took me a bit (15 whole seconds of reading, the horror!) to find the link to the spec.
> 
> I'd prefer to link straight to the spec and have links to test suite and other things from the spec. My concern is hitting a page that doesn't look "official" when going to the ED. That said, I don't feel strongly about this, so happy to defer to the Editors and Chairs (and/or WG) on the best set up. Each WG has its own culture. :)

The practice I've seen lately is to put the ED in a /spec structure, and separate out /tests and so forth. It does mean that https://w3c.github.io/rdf-canon/ takes you to a page (README) that links to the spec. The "Latest editor's draft" link on the /TR version can probably be tweaked to take you straight to https://w3c.github.io/rdf-canon/spec.

gkellogg (changes requested):
Use `VALIDATE_LINKS` and `VALIDATE_MARKUP` from ci.yaml, ~~and remove ci.yaml~~ as being redundant.

.github/workflows/auto-publish.yml
```diff
@@ -0,0 +1,23 @@
...
+    paths: ["spec/**"]
+  push:
+    branches: [main]
+    paths: ["spec/**"]
+jobs:
+  main:
+    name: Echidna Auto-publish WD
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
```
gkellogg, line 14:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/auto-publish.yml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: w3c/spec-prod@v2
+        with:
+          TOOLCHAIN: respec
+          SOURCE: spec/index.html
+          W3C_ECHIDNA_TOKEN: ${{ secrets.ECHIDNA_TOKEN }}
```
gkellogg, line 19:
```suggestion
          W3C_ECHIDNA_TOKEN: ${{ secrets.ECHIDNA_TOKEN }}
          VALIDATE_LINKS: false
          VALIDATE_MARKUP: true
```

gkellogg:
> I'd prefer to link straight to the spec and have links to test suite and other things from the spec. My concern is hitting a page that doesn't look "official" when going to the ED. That said, I don't feel strongly about this, so happy to defer to the Editors and Chairs (and/or WG) on the best set up. Each WG has its own culture. :)

Fixed in #51.

yamdan (approved):

PR #53, opened 2022-12-07
Removed the _: strings from the algorithms and the examples

### LINKED ISSUE #46: [Editorial] Discrepancy in the way we refer to Blank Node IDs
There is a slight discrepancy between the RDF terminology and the terminology used in our specification document for blank node identifiers. 

The description of the algorithm, all the examples, and probably all the implementations, systematically use the format `_:xyz` for blank node identifiers. This is also significant, e.g., in the Hash Related Blank Node algorithm that uses the blank node id-s as part of a path that gets hashed at a later step. 

This is not in line with the [RDF Concepts and Abstract Syntax](https://www.w3.org/TR/rdf11-concepts/#section-blank-nodes) spec, that speaks of abstract blank node id-s in general. The usage of the `_:` _is only a serialization artifact for Turtle, N-Triple, or JSON-LD_. Note that, in RDF/XML, blank nodes are identified using the `rdf:nodeID` attribute, whose value is an arbitrary string; no `_:` there.

The reason why this matters is because RDF implementations, I presume, usually rely on the RDF spec and do _not_ use `_:`, except for parsing and serialization to Turtle, N-Triple, or JSON-LD. E.g., in `rdf-js` a blank node is created through `DataFactory.BlankNode('abc')` and, more importantly for our use, when the id of a blank node is retrieved via `bnode.value()`, the identifier being returned is without the `_:`. The same happens for Python's RDFLib. This type of discrepancy can be a source of implementation bugs that can be difficult to trace. 

I presume that, at this moment, it is too late to change this, because we do not want to make existing implementations fail on something like that. But it may be worth adding a note somewhere to warn the users about this potential issue and also make it clear in the implementation (e.g., in step 4 of the Hash Related Blank Node algorithm) that the `_:` _must_ be presented in the constructed string being hashed. This type of warning might be helpful for implementers...

cc @pchampin @gkellogg @davidlehn @dlongley

iherman:
This is an attempt to fix #46.

The algorithm descriptions, specifications, the examples, etc, abound with the term "identifier", so I opted for keeping this term for the blank node identifier in the sense of the RDF specs. This means that the change "only" consists of removing the `_:` string from everywhere, except the turtle code in the example blocks. I could not touch the SVG files, because I do not have access to the originals (and the SVG files themselves do not include any string).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/53.html" title="Last updated on Dec 12, 2022, 7:38 PM UTC (7a8b2fe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/53/5299b72...7a8b2fe.html" title="Last updated on Dec 12, 2022, 7:38 PM UTC (7a8b2fe)">Diff</a>


gkellogg:
I presume that you checked to see if it produces the same output for the tests? My impression is that it would be the same, as removing a common prefix shouldn't affect ordering.

spec/index.html
```diff
@@ -310,13 +310,17 @@ <h2>Terminology</h2>
...
+      as specified by [[!RDF11-CONCEPTS]]. In short, it is a <a>string</a> that is used as an identifier for a
       <a>blank node</a>. <a>Blank node identifiers</a>
       are typically implementation-specific local identifiers; this document
       specifies an algorithm for deterministically specifying them.
       A blank node not given an explicit identifier is called an "unlabeled blank node."
       Additionally, blank nodes may be explicitly labeled with an identifier.</dd>
+    <dd>
+      Concrete syntaxes, like [[Turtle]] or [[N-Quads]], prepend the blank node identifiers with the <code>_:</code> string
+      to differentiate them from other nodes in graph. This affects the 
+      canonicalization algorithm, which is based on calculating a hash over the representations <a>quads</a> in this format.
```
TallTed, line 322:
```suggestion
      Concrete syntaxes, like [[Turtle]] or [[N-Quads]], prepend blank node identifiers with the <code>_:</code> string
      to differentiate them from other nodes in the graph. This affects the 
      canonicalization algorithm, which is based on calculating a hash over the representations of <a>quads</a> in this format.
```

iherman:
> I presume that you checked to see if it produces the same output for the tests? My impression is that it would be the same, as removing a common prefix shouldn't affect ordering.

I still have to go through the full test suite, which I have not done yet. I have checked with the examples in the spec and there each step produced identical results with the old setup.

But... even if there are discrepancies, I believe the approach without _: is the right or, shall we say, the "clean" one for the algorithm. The current text is conceptually messy, imho.

dlongley:
Without looking at this more closely ... the `_:` prefix was also used as a delimiter in some places, so removing it could cause some paths to be confused with others.

iherman:
> Without looking at this more closely ... the `_:` prefix was also used as a delimiter in some places, so removing it could cause some paths to be confused with others. Or, at the very least, some analysis would have to be done to rule that out. I'm -1 to anything that changes the output over this since it seems very minor to me. +1 to anything we can do that is helpful but doesn't change the output.

Would it be possible to locate those steps explicitly, @dlongley ? I suspect it is in step (4) in the [Hash Related Blank Node](https://www.w3.org/TR/rdf-canon/#hash-related-algorithm). Anywhere else? Because it is perfectly fine to change that step and add the `_:` to _identifier_ explicitly. 

I must admit that, now that I have done the change mentally :-), I find the current algorithmic description pretty confusing. It sounds as if the algorithm was done on nquads and not on the RDF Graph in the abstract sense.

dlongley:
@iherman,

> Would it be possible to locate those steps explicitly, @dlongley ? I suspect it is in step (4) in the [Hash Related Blank Node](https://www.w3.org/TR/rdf-canon/#hash-related-algorithm). Anywhere else?

Yes, step 4 there needs it, so does the [Hash N-Degree Quads](https://www.w3.org/TR/rdf-canon/#hash-nd-quads-algorithm) in every step that appends an identifier, which seem to be steps: 5.4.4.1, 5.4.4.2 (both substeps), and 5.4.5.2.

I found those steps by searching for the text "append" and using any instances where a blank node identifier was appended. In all such cases, the `_:` prefix is important as a delimiter.

dlongley:
@iherman,

> I must admit that, now that I have done the change mentally :-), I find the current algorithmic description pretty confusing. It sounds as if the algorithm was done on nquads and not on the RDF Graph in the abstract sense.

I didn't quite follow this comment. It is true that the RDF graph needs to be serialized in some way so it can be hashed to enable hash digest comparison.

iherman:
> On 8 Dec 2022, at 18:27, Dave Longley ***@***.***> wrote:
> 
> 
> @iherman <https://github.com/iherman>,
> 
> Would it be possible to locate those steps explicitly, @dlongley <https://github.com/dlongley> ? I suspect it is in step (4) in the Hash Related Blank Node <https://www.w3.org/TR/rdf-canon/#hash-related-algorithm>. Anywhere else?
> 
> Yes, step 4 there needs it, so does the Hash N-Degree Quads <https://www.w3.org/TR/rdf-canon/#hash-nd-quads-algorithm> in every step that appends an identifier, which seem to be steps: 5.4.4.1, 5.4.4.2 (both substeps), and 5.4.5.2.
> 
> I found those steps by searching for the text "append" and using any instances where a blank node identifier was appended. In all such cases, the _: prefix is important as a delimiter.
> 
I am happy to modify the PR to take these into account.

iherman:
> @iherman,
> 
> > I must admit that, now that I have done the change mentally :-), I find the current algorithmic description pretty confusing. It sounds as if the algorithm was done on nquads and not on the RDF Graph in the abstract sense.
> 
> I didn't quite follow this comment. It is true that the RDF graph needs to be serialized in some way so it can be hashed to enable hash digest comparison.

Yes, and that is a perfectly good reason to prepend the identifier with `_:` for all those cases; hashing means some sort of serialization of the terms.

But, e.g., when I issue a new ID, when I prepend it with `c14n` or `b`, or anything else, I operate on the abstract graph. Even in an implementation, I operate on the, say, the class representing a bnode (or other types of RDF terms), not on their serialization. In those circumstances, using the `_:` should be out of space, because we are _not_ operating on the serialization of the RDF term.

gkellogg:
> But, e.g., when I issue a new ID, when I prepend it with `c14n` or `b`, or anything else, I operate on the abstract graph. Even in an implementation, I operate on the, say, class representing a bnode (or other types of RDF terms), not on their serialization. In those circumstances, using the `_:` should be out of space, because we are _not_ operating on the serialization of the RDF term.

Yes, I think we can separate these, the the Blank Node identifier is prefixed by `c14n` or `b`, but serialized with by additionally prepending `_:`. This keeps a separating between the identifiers (or labels) and their serialized form, as required by the algorithm.

iherman:
> @iherman,
> 
> > Would it be possible to locate those steps explicitly, @dlongley ? I suspect it is in step (4) in the [Hash Related Blank Node](https://www.w3.org/TR/rdf-canon/#hash-related-algorithm). Anywhere else?
> 
> Yes, step 4 there needs it, so does the [Hash N-Degree Quads](https://www.w3.org/TR/rdf-canon/#hash-nd-quads-algorithm) in every step that appends an identifier, which seem to be steps: 5.4.4.1, 5.4.4.2 (both substeps), and 5.4.5.2.
> 
> I found those steps by searching for the text "append" and using any instances where a blank node identifier was appended. In all such cases, the `_:` prefix is important as a delimiter. It may be better to say that whenever a blank node identifier (or "label" should that end up being the language used) is appended to a string, it must include that prefix.

- I have taken care of these except for 5.4.4.2.1. I believe there is no change there: the 'append' term refers to the fact that the bnode id must be added to the _recursion list_. I believe that must stay in its abstract form.
- I have updated example 6 and example 7 (hopefully I got them right :-)
- I have added an explanation to point (4) in §4.8.3, and I have repeated it in §4.9.3 point 5.4.4

dlongley (approved):
Ok, this looks like it will preserve the places where the identifiers were serialized in the algorithm but remove the `_:` serialization prefix where they are discussed in the abstract sense. Thanks!

Note: I agree with you @iherman for that one step where the identifier is added to a recursion list (not to a path) that it shouldn't have the `_:` prefix there.

spec/index.html
```diff
@@ -595,12 +599,12 @@ <h3>Examples</h3>
           </tr></thead>
           <tbody>
             <tr>
-              <td>`_:e0`</td>
-              <td>`:p :q _:e0 .`<br/>`_:e0 :s :u .`</td>
+              <td>`e0`</td>
+              <td>`:p :q e0 .`<br/>`e0 :s :u .`</td>
```
gkellogg, line 603: As this looks like a serialization of the quad, blank nodes should either be represented using the `_:` prefix, or they should be otherwise distinguished, possibly using `<var>`, `<em>` or similar, which may require some explanation of nomenclature. Easiest is probably to stick with `_:e0` when presenting a quad unless you think it's key that we stick with the abstract syntax, in which case we probably need to define our notation.

gkellogg (changes requested):
This looks really good. We need to clarify how we present quads in examples, either with a `_:` or using some other distinguishing characteristic. For most readers, seeing the quads in N-Quads format may be the most natural.

gkellogg:
I did the other part of #46 and clarified that input nodes are just "blank nodes", not using the term "label" or "identifier", and that in example steps, they result in either "canonical" or "temporary" identifiers.

Fixes #46.

dlongley (commented):
I think these changes are internally consistent with one another, but I don't know how much easier it is to understand now (or if it's potentially harder?).

I can only speak for myself, but I prefer names like "original" that help provide some context around what the original label was when implementing and trying to understand and debug things. Obviously implementers can still do whatever they want for naming things.

The same goes for reading the spec examples where `_:` was previously present and could be used to pick up on where the blank nodes are.

I don't intend to hold these changes up in any way, however, if most people feel that this consistency and presentation is better -- I may be the odd one out.

gkellogg:
> I think these changes are internally consistent with one another, but I don't know how much easier it is to understand now (or if it's potentially harder?).
> 
> I can only speak for myself, but I prefer names like "original" that help provide some context around what the original label was when implementing and trying to understand and debug things. Obviously implementers can still do whatever they want for naming things.

I wanted to clarify that the input blank node (original) doesn't need to discuss an identifier, and that, say _e1_ is how we refer to the blank node, not necessarily some assigned or presumed identifier, which it may not have. But, we can continue to use "original" rather than "blank node" in these cases.

> The same goes for reading the spec examples where `_:` was previously present and could be used to pick up on where the blank nodes are.
> 
> I don't intend to hold these changes up in any way, however, if most people feel that this consistency and presentation is better -- I may be the odd one out.

It's certainly worth a wider review. Omitting the `_:` when discussing blank node identifiers is consistent with the description in Concepts. Perhaps more narrative in the use of `_:` when serializing would help?

gkellogg:
> I wanted to clarify that the input blank node (original) doesn't need to discuss an identifier, and that, say _e1_ is how we refer to the blank node, not necessarily some assigned or presumed identifier, which it may not have. But, we can continue to use "original" rather than "blank node" in these cases.
> 
> >

Looking at the examples, I'm not sure that "original" really makes sense, as we are talking about blank nodes, implying that they are the blank nodes from the input dataset. Perhaps "original blank node" would work better. I'll wait for some more opinions before making further changes.

spec/index.html
iherman: I must admit I hesitated on this. The problem that I can see is that in the explanation text the bare id is used, as it should be, hence my choice of removing the infamous `_:`. 

Actually, it is simple to distinguish the blank nodes in all those places: they are the only symbols without the `:` character at the beginning.

iherman:
> > I wanted to clarify that the input blank node (original) doesn't need to discuss an identifier, and that, say _e1_ is how we refer to the blank node, not necessarily some assigned or presumed identifier, which it may not have. But, we can continue to use "original" rather than "blank node" in these cases.
> > >
> 
> Looking at the examples, I'm not sure that "original" really makes sense, as we are talking about blank nodes, implying that they are the blank nodes from the input dataset. Perhaps "original blank node" would work better. I'll wait for some more opinions before making further changes.

I did not find the text you are referring to look at. But the way you formulate this is not really right. Sorry to be pedantic, but I believe the blank nodes appearing in the original graph as well as the canonical one _are the same_, only their identification are different. Talking about "original blank node", as opposed to canonical once, is wrong...

(Calling on @pchampin to tell me if I am right per RDF)

gkellogg:
> > > I wanted to clarify that the input blank node (original) doesn't need to discuss an identifier, and that, say _e1_ is how we refer to the blank node, not necessarily some assigned or presumed identifier, which it may not have. But, we can continue to use "original" rather than "blank node" in these cases.
> > > >
> > 
> > 
> > Looking at the examples, I'm not sure that "original" really makes sense, as we are talking about blank nodes, implying that they are the blank nodes from the input dataset. Perhaps "original blank node" would work better. I'll wait for some more opinions before making further changes.
> 
> I did not find the text you are referring to look at. But the way you formulate this is not really right. Sorry to be pedantic, but I believe the blank nodes appearing in the original graph as well as the canonical one _are the same_, only their identification are different. Talking about "original blank node", as opposed to canonical once, is wrong...
> 
> (Calling on @pchampin to tell me if I am right per RDF)

Yes, you’re correct that they are the same blank node, just with different identifiers. Updates I made were to get away from using “identifier” or “label” when referring to the input, and just use the identifier from the example as its stand-in. A blank node in the input may have no assigned identifier.

generally, the text in the algorithms and examples (where tables are involved) describe one side using “blank node” and do not reference an identifier, so using that term as a column heading didn’t make much sense. In the case of “original”, this may be different, but still implies that the blank node from the input has an identifier, which it may not, while in the normalized output, it always will have a canonical identifier.

spec/index.html
gkellogg: In the updated diagrams, the identifiers are italicized, so maybe that would help. I can do that in an update.

dlongley:
> Updates I made were to get away from using “identifier” or “label” when referring to the input, and just use the identifier from the example as its stand-in. A blank node in the input may have no assigned identifier.

While it's true a blank node in the input may have no assigned identifier, I wonder if someone implementing the algorithm would benefit more from having an original label from which to work off of. Especially if any of these examples end up as tests in the test suite.

The test suite, at least today, parses the input from a serialization where the blank nodes would have an identifier that is helpful to use as a reference point for understanding what's going on. I can imagine wanting to run those tests specifically and walk through what's in the spec and compare it against one's implementation and the examples. Using abstract blank nodes there, to me, seems less helpful than using "original identifiers". This is what I meant in my comment around "original", but, again, others could see it another way -- and maybe it's easy enough to map these things either way so it doesn't matter that much.

spec/index.html
```diff
@@ -614,38 +622,38 @@ <h3>Examples</h3>
...
           blank node which has a unique hash:</p>
 
-        <table id="ex-ca-unique-1d-labels"
+        <table id="ex-ca-unique-1d-identifiers"
                class="hash-related numbered">
-          <caption>Canonical labels for unique hashes</caption>
+          <caption>Canonical identifiers for blank nodes with unique hashes</caption>
           <thead><tr>
-            <th><var>label</var></th>
+            <th><var>blank node</var></th>
```
gkellogg, line 646: Here, I think the use of "blank node" vs "label" is justified, as the introducing paragraph uses the term "blank node", so it should not be confusing.

spec/index.html
```diff
@@ -703,60 +711,60 @@ <h3>Examples</h3>
...
+          (`e0`, `e1`, `e2`, and `e3`)
           to populate <a>blank node to quads map</a>:</p>
 
         <table id="ex-ca-shared-bn-quads-map"
                class="hash-related numbered">
           <caption>Blank node to quads map for shared hashes</caption>
           <thead><tr>
-            <th><var>identifier</var></th>
+            <th><var>blank node</var></th>
             <th><var>Q</var></th>
```
gkellogg, line 722: Same here, its a "blank node to quads map", so showing a "blank node" column corresponds.

spec/index.html
```diff
@@ -795,18 +803,18 @@ <h3>Examples</h3>
...
+          <caption>Result for `e0`</caption>
           <tbody>
             <tr>
               <td><var>temporary issuer</var> mappings</td>
               <td>
                 <table class="hash-related">
                   <thead><tr>
-                    <th><var>original</var></th>
-                    <th><var>canonical</var></th>
+                    <th><var>blank node</var></th>
```
gkellogg, line 813: This seems to be the main area of confusion. Perhaps "original identifier" is more clear here?

gkellogg:
To make this change, I think we need full consensus. From what I understand, @dlongley feels that the use of the `_:` nomenclature is more natural for developers to understand, and the fact that more formally these are a serialization artifact, and not part of the abstract syntax, is not helpful for understanding.

@iherman rightly points out that it is just _wrong_ to talk about a blank node identifier abstractly using the `_:` syntactic artifact, and it is semantically clearer to use just the pure identifier.

My main take was to try to separate the terms "label" and "identifier", and I think the two concepts are reasonably clear in the text of this PR.

I'd like to get opinions in particular from @pchampin and @yamdan as well as anyone else that would like to chime in.

iherman:
> 
> While it's true a blank node in the input may have no assigned identifier, I wonder if someone implementing the algorithm would benefit more from having an original label from which to work off of. Especially if any of these examples end up as tests in the test suite.
> 
> The test suite, at least today, parses the input from a serialization where the blank nodes would have an identifier that is helpful to use as a reference point for understanding what's going on. I can imagine wanting to run those tests specifically and walk through what's in the spec and compare it against one's implementation and the examples. Using abstract blank nodes there, to me, seems less helpful than using "original identifiers". This is what I meant in my comment around "original", but, again, others could see it another way -- and maybe it's easy enough to map these things either way so it doesn't matter that much.

@dlongley actually, what you say is not always possible, it relies on some assumption on the implementation. In my case, I rely on an external RDF library that parses the Turtle/TriG input and returns a set of quad objects, including blank node objects. Although the blank node objects do have an identifier, that identifier is not identical to the identifier in the original source. And that is perfectly acceptable per the RDF standard. What exactly those identifiers are, depend on the parser.

(The quads -> nquads serialization _usually_ reuse the internal Id of the blank nodes, but I do not think this is defined per standard. Serializers _may_ choose to change those, too, and implementations must be careful about that...)

iherman:
> @iherman rightly points out that it is just _wrong_ to talk about a blank node identifier abstractly using the `_:` syntactic artifact, and it is semantically clearer to use just the pure identifier.

Let me add that it is not only wrong semantically, but it is also disturbing when one implements the algorithm on top of an off-the-shelf RDF library. Such a library usually does not keep the `_:` prefix in the bnode identifiers. The proposed changes make this difference more clearly and therefore reflects how the implementation should be done when embedded in such an RDF library. Ie, the changes are very much practical, it is not only a matter of purity.

iherman:
My last remarks made me realize... the spec should say somewhere that the nquad serialization _MUST_ retain the blank node identifiers except the pretending of `:_`. This is essential both for the algorithm itself, and when we will talk about hashing the canonical graph.

This seems obvious, and I suspect all RDF libraries do it that way. But _I don't think any of the nquad specification require this,_ so we have to make it sure in our document(s).

Editorially, it is probably the best to put a generic text somewhere instead of complicating the algorithmic description itself.

gkellogg:
> My last remarks made me realize... the spec should say somewhere that the nquad serialization _MUST_ retain the blank node identifiers except the pretending of `:_`. This is essential both for the algorithm itself, and when we will talk about hashing the canonical graph.
> 
> This seems obvious, and I suspect all RDF libraries do it that way. But _I don't think any of the nquad specification require this,_ so we have to make it sure in our document(s).
> 
> Editorially, it is probably the best to put a generic text somewhere instead of complicating the algorithmic description itself.

Thus far, we are pretty explicit about how the blank node is rendered in the canonical n-quads output. When we go to describing a serialization of a normalized dataset into canonical n-quads form, we'll need to be explicit about rendering the canonical identifiers in the output.

pchampin (approved):
Sorry for the late reaction. I agree with this change. As @iherman this is not just about semantic purity, but also about acknowledging that the `_:` is not part of the bnode identifier in all implementations.

gkellogg (approved):

marcelotto:
Am I the only getting failing tests (12 tests of the test suite) when applying these changes?

It seems to change the behaviour of at least the _Hash Related Blank Node_ algorithm, where the `_:` is now prepended to the hashes produced by the application of the _Hash First Degree Quads_ algorithm in step 1.

gkellogg:
Those change shouldn't affect the previous way the algorithms work, just nomenclature. Effectively, blank nodes will all have a form such as `_:c14n1` (or whatever the algorithm states), its just that formally, the _identifier_ is `c14n1`, and the `_:` is added during the serialization process.

marcelotto:
I understand that this shouldn't affect the identifier created by an issuer, but in step 1 of the _Hash Related Blank Node_ algorithm also a hash might be assigned to `identifier`.

> 1) Set the identifier to use for related, preferring first [...] and last, if necessary, **the result of the _Hash First Degree Quads algorithm_**, passing related.

Prepending a `_:` to this hash in step 4 will change the `input` value passed to the hash algorithm and with that, the resulting hash of _Hash Related Blank Node_ compared to the previous version. 

As I've said, I tested it and I'm getting different results (12 failing tests of the test suite). That's why I'm asking if someone else also tested these changes and experienced the same result.

gkellogg:
> Prepending a `_:` to `identifier` when it contains a hash in step 4 will change the `input` value passed to the hash algorithm and with that, the resulting hash of _Hash Related Blank Node_ compared to the previous version.

Yes, you are correct that prepending `_:` to the result of the Hash First Degree Quads algorithm would be incorrect,  and this is not accounted for. I'll create a new issue for this, and follow-on PR.

PR #55, opened 2022-12-10
 Use ReSpec autolinks for links to externally defined terms

dontcallmedom:
Also, add index of locally and externally defined terms and fix link to editors draft

***
<a href="https://pr-preview.s3.amazonaws.com/dontcallmedom/rdf-canon/pull/55.html" title="Last updated on Dec 12, 2022, 7:22 PM UTC (372de39)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/55/3f09a10...dontcallmedom:372de39.html" title="Last updated on Dec 12, 2022, 7:22 PM UTC (372de39)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -21,7 +21,7 @@
     // publishDate:  "2009-08-06",
 
     github: "https://github.com/w3c/rdf-canon/",
-
+    edDraftURI: "https://w3c.github.io/rdf-canon/spec/",
     // editors, add as many as you like
     // only "name" is required
     editors:  [
--- a/spec/index.html
+++ b/spec/index.html
@@ -63,9 +63,10 @@
     // name of the 
     group: "rch",
     testSuiteURI: "https://w3c.github.io/rdf-canon/tests/",
-    xref: ["rdf11-concepts", "rdf11-mt"],
+    xref: ["rdf11-concepts", "infra"],
     doJsonLd:     true,
-    wgPublicList: "public-rch-wg"
+    wgPublicList: "public-rch-wg",
+    lint: { "informative-dfn": false }
   };
 //]]>
 </script>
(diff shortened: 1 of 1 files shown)


dlongley (approved):
LGTM from looking at the links (with appear to be properly auto-generated) in the preview.

spec/index.html
```diff
@@ -271,30 +272,6 @@ <h2>Terminology</h2>
...
-      as specified by [[!RDF11-CONCEPTS]].</dd>
-    <dt><dfn data-lt="rdf graph|rdf graphs|graph|graphs">RDF graph</dfn></dt>
-    <dd>An <a data-cite="RDF11-CONCEPTS#dfn-rdf-graph">RDF graph</a>
-      as specified by [[!RDF11-CONCEPTS]].</dd>
-    <dt><dfn data-lt="graph name|graph names">graph name</dfn></dt>
-    <dd>A <a data-cite="RDF11-CONCEPTS#dfn-graph-name">graph name</a>
-      as specified by [[!RDF11-CONCEPTS]].</dd>
-    <dt><dfn>default graph</dfn></dt>
-    <dd>The <a data-cite="RDF11-CONCEPTS#dfn-default-graph">default graph</a>
-      as specified by [[!RDF11-CONCEPTS]].</dd>
```
gkellogg, line 297: Not sure we want to lose all these entries, although they definitely should not be exported terms and should use the `data-cite` mechanism. For someone not relying on on hyperlinks, having a basic definition of the terminology used in the spec in one place, whether imported or exported, is useful.

spec/index.html
```diff
@@ -898,7 +875,7 @@ <h3>Algorithm</h3>
           <ol>
             <li id="ca-2-1">For each <a>blank node</a> that is a component of <var>Q</var>,
               add a reference to <var>Q</var> from the
-              <a data-cite="INFRA#map-entry">map entry</a> for the
+              [= map/entry | map entry =] for the
```
gkellogg, line 878: For whatever reason, I've been reluctant to use the `[= ... =]` form, but I suppose I need to catch up with the times :).

In this case, the `data-cite="INFRA#map-entry"` does seem to resolve to the proper location (https://infra.spec.whatwg.org/#map-entry), and I'm not sure why, or how I would figure out that "map/entry" is the proper way to reference the term.

gkellogg: Perhaps that can be inferred from the "#map-entry" fragment identifier, which implies a hierarchical relationship.

gkellogg (commented):
Thanks for the general improvements! If it's reasonable to keep the definitions for some of the imported terms (as cited references), I'll update them to use `data-cite`, unless you had some other reason for removing them to consider.

spec/index.html
dontcallmedom: except for IRI, none of these provide anything more than just links to their original definition, so I don't think removing them remove any local context; but naturally, I'm happy to reinstate them as data-cite if you prefer.

spec/index.html
dontcallmedom: To find how to refer to an existing term, https://respec.org/xref/?term= is the right tool; the syntax `map/entry` comes from the fact that the term `entry` is defined as a concept in the context of `map`.

spec/index.html
gkellogg: Thanks, indeed, they provide only a local (possibly contextual) repetition of a term defined elsewhere. This comes down to a style point, more than anything. I think it best to keep them in place, and then later consider pruning back such terminology lists.

As this is out of your repo, it would be great if you'd put them back and use `<dfn data-cite="...">` so that they are just fancy references.

I'll create an issue about the need for local term definitions for imported terms.

iherman:
The main advantage I see is that we do not have to care about references that may change in future; all changes are done automatically. +1 to do this.

Thanks @dontcallmedom

PR #45, opened 2022-11-26
More examples

gkellogg:
This creates an example for the Hash N-Degree Quads algorithm, using the "Shared hashes" example.

~~@dlongley note the editor's note added to canonicalization, which you might comment on.~~

~~This needs a similar example run through from the Canonicalization algorithm, which can just use the outputs of this example.~~

Fixes #1.
Fixes #34.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/45.html" title="Last updated on Dec 2, 2022, 11:51 PM UTC (9955102)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/45/cc69323...9955102.html" title="Last updated on Dec 2, 2022, 11:51 PM UTC (9955102)">Diff</a>

PR #56, opened 2022-12-14
Renamed `_:b*` types names to `_:e*` in the tests

iherman:
In line with a [similar commit](https://github.com/w3c/rdf-canon/pull/45/commits/d57bbd9575a5b47652c02b94f2c7f76ed7b85be3) I changed all the occurences of `_:b*` blank node names to `_:e* in all the tests.

Reminder: in [Step 5.2.2.](https://www.w3.org/TR/rdf-canon/#ca-5-2-2) we create a new, temporary ID issuer with the prefix `b`. If the original nquads use this prefix (and the parser keeps these prefixes) it leads to major confusions when looking at debug logs or examples. Hence the aforementioned commit for the examples in the spec, and this PR does the same for all official tests.


iherman:
Actually, test No. 59 has changed more thoroughly. The original test used fairly cryptic and difficult-to-use blank nodes; this PR aligns them to the style of the other tests.

dlongley:
It's probably important to continue to have at least some tests that use `_:b` as a prefix to ensure that an implementation doesn't have a problem because of it.

gkellogg (approved):

dlongley (approved):
+1 to duplicating some tests to use `_:b`.

marcelotto:
It seems it was forgotten to add test 63 to the manifests.

PR #57, opened 2022-12-15
Minor editorial knit: "canonical issuer' or just 'issuer'

iherman:
In 5.4.4.1 of the Hash N-Degree Quads algorithm, it said

> If a canonical identifier has been issued…

~I propose to remove the term "canonical". It may mislead the reader (it misled me) to use the canonical issuer of the canonicalization state, which is (I believe) not the case here.~

The term "issuer" by itself misleads the reader (it misled me) to use the local, temporary issuer, rather than the canonical one.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/57.html" title="Last updated on Dec 16, 2022, 11:09 PM UTC (5fd2500)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/57/5b1e80f...5fd2500.html" title="Last updated on Dec 16, 2022, 11:09 PM UTC (5fd2500)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2036,7 +2036,7 @@
                   recursively processed by this algorithm.</li>
                 <li id="hndq-5-4-4">For each <var>related</var> in <var>p</var>:
                   <ol>
-                    <li id="hndq-5-4-4-1">If a canonical identifier has been issued for
+                    <li id="hndq-5-4-4-1">If an identifier has been issued for
                       <var>related</var> by <var>issuer</var>, append the string <code>_:</code>, followed by 
                       <var>related</var>, to <var>path</var>.</li>
                     <li id="hndq-5-4-4-2">Otherwise:
--- a/spec/index.html
+++ b/spec/index.html
@@ -2038,7 +2038,14 @@
                   <ol>
                     <li id="hndq-5-4-4-1">If a canonical identifier has been issued for
                       <var>related</var> by <a>canonical issuer</a>, append the string <code>_:</code>, followed by 
-                      <var>related</var>, to <var>path</var>.</li>
+                      the canonical identifier for <var>related</var>, to <var>path</var>.
+                      <details><summary>Explanation</summary>
+                        <p>A canonical identifier may have been generated before calling this algorithm,
+                            if it was issued from an earlier call to <a href="#hash-1d-quads">Hash First Degree Quads algorithm</a>.
+                            There is no reason to recurse and apply the algorithm to any related blank node that has already been assigned a canonical identifier.
+                             Furthermore, using the canonical identifier also further distinguishes it from any temporary identifier, allowing for even greater efficiency in finding the chosen path.</p>
+                      </details>
+                    </li>
                     <li id="hndq-5-4-4-2">Otherwise:
                       <ol>
                         <li id="hndq-5-4-4-2-1">If <i>issuer copy</i> has not issued
--- a/spec/index.html
+++ b/spec/index.html
@@ -2036,8 +2036,8 @@
                   recursively processed by this algorithm.</li>
                 <li id="hndq-5-4-4">For each <var>related</var> in <var>p</var>:
                   <ol>
-                    <li id="hndq-5-4-4-1">If an identifier has been issued for
-                      <var>related</var> by <var>issuer</var>, append the string <code>_:</code>, followed by 
+                    <li id="hndq-5-4-4-1">If a canonical identifier has been issued for
+                      <var>related</var> by <a>canonical issuer</a>, append the string <code>_:</code>, followed by 
                       <var>related</var>, to <var>path</var>.</li>
                     <li id="hndq-5-4-4-2">Otherwise:
                       <ol>


dlongley:
I'll have to check on this, but my initial reaction was that it was actually correct and the canonical issuer should be used here *if* it has already issued an identifier for `related`.

dlongley:
Yes, it was correct -- and here's what DB's JavaScript implementation does:

```js
// 5.4.4.1) If a canonical identifier has been issued for
// related, append it to path.
if(this.canonicalIssuer.hasId(related)) {
  path += this.canonicalIssuer.getId(related);
} else {
  // 5.4.4.2) Otherwise:
  // 5.4.4.2.1) If issuer copy has not issued an identifier for
  // related, append related to recursion list.
  if(!issuerCopy.hasId(related)) {
    recursionList.push(related);
  }
  // 5.4.4.2.2) Use the Issue Identifier algorithm, passing
  // issuer copy and related and append the result to path.
  path += issuerCopy.getId(related);
}
```

https://github.com/digitalbazaar/rdf-canonize/blob/main/lib/URDNA2015.js#L306-L320

dlongley (changes requested):
Requesting that we don't merge this because it seems correct to me. Maybe the text that's being changed here should remain, but we should also be more clear about which `issuer` is being checked / used.

spec/index.html
```diff
@@ -2036,7 +2036,7 @@ <h3>Algorithm</h3>
                   recursively processed by this algorithm.</li>
                 <li id="hndq-5-4-4">For each <var>related</var> in <var>p</var>:
                   <ol>
-                    <li id="hndq-5-4-4-1">If a canonical identifier has been issued for
+                    <li id="hndq-5-4-4-1">If an identifier has been issued for
                       <var>related</var> by <var>issuer</var>, append the string <code>_:</code>, followed by
```
dlongley, line 2040:
```suggestion
                    <li id="hndq-5-4-4-1">If a canonical identifier has been issued for
                      <var>related</var> by <a>canonical issuer</a>, append the string <code>_:</code>, followed by 
```

gkellogg:
My code is similar to @dlongley's:

```ruby
permutation.each do |related|
  if canonical_issuer.identifier(related)
    path << canonical_issuer.issue_identifier(related)
  else
    recursion_list << related if !issuer_copy.identifier(related)
    path << issuer_copy.issue_identifier(related)
  end

  # Skip to the next permutation if chosen path isn't empty and the path is greater than the chosen path
  break if !chosen_path.empty? && path.length >= chosen_path.length
end
```

gkellogg:
@dlongley, can you suggest an Explaination/details that we would add to 5.4.4.1 to clarify why this is so?

The examples for step 5 could also use a phrase to describe why the canonical identifier is used here.

spec/index.html
```diff
@@ -2036,7 +2036,7 @@ <h3>Algorithm</h3>
                   recursively processed by this algorithm.</li>
                 <li id="hndq-5-4-4">For each <var>related</var> in <var>p</var>:
                   <ol>
-                    <li id="hndq-5-4-4-1">If a canonical identifier has been issued for
+                    <li id="hndq-5-4-4-1">If an identifier has been issued for
                       <var>related</var> by <var>issuer</var>, append the string <code>_:</code>, followed by 
                       <var>related</var>, to <var>path</var>.</li>
```
gkellogg, line 2041:
```suggestion
                      <var>related</var>, to <var>path</var>.
                      <details><summary>Explanation</summary>
                        <p>A canonical identifier may have been generated before calling this algorithm,
                            if it was issued from an earlier call to <a href="#hash-1d-quads">Hash First Degree Quads algorithm</a>.</p>
                      </details>
                    </li>

**Review by gkellogg — CHANGES_REQUESTED** [review]
I think the example text is okay, but the details I added may make this more clear.

File: spec/index.html (hunk 2 above)
Comment (dlongley) [reply spec/index.html:2041]: > can you suggest an Explanation/details that we would add to 5.4.4.1 to clarify why this is so?

I think that suggested explanation is fine. If you want any other details: there's no reason to recurse and apply the algorithm to any related blank node that has already been assigned a canonical identifier. Furthermore, using the canonical identifier also further distinguishes it from any temporary identifier, allowing for even greater efficiency in finding the chosen path.

File: spec/index.html (hunk 1 above)
Comment (iherman) [reply spec/index.html:2040]: Ah! So there was indeed a discrepancy, but just not where I thought it was...

**Comment by iherman** [comment]
> Requesting that we don't merge this because it seems correct to me. Maybe the text that's being changed here should remain, but we should also be more clear about which `issuer` is being checked / used.

I accepted the change, so the discrepancy is settled, just the other way round than what the PR started with...

File: spec/index.html [hunk 3]
```diff
@@ -2037,7 +2037,7 @@ <h3>Algorithm</h3>
                 <li id="hndq-5-4-4">For each <var>related</var> in <var>p</var>:
                   <ol>
                     <li id="hndq-5-4-4-1">If a canonical identifier has been issued for
-                      <var>related</var> by <var>issuer</var>, append the string <code>_:</code>, followed by 
+                      <var>related</var> by <a>canonical issuer</a>, append the string <code>_:</code>, followed by 
                       <var>related</var>, to <var>path</var>.</li>
```
iherman, line 2041:
```suggestion
                      <var>related</var> by <a>canonical issuer</a>, append the string <code>_:</code>, followed by 
                      the canonical identifier for <var>related</var>, to <var>path</var>.</li>
```

(The original text suggests that the value of _related_ should be appended to _path_...)

TallTed:
s/editorial [`knit`](https://www.google.com/search?q=define%3Aknit)/editorial [`nit`](https://www.techopedia.com/definition/31651/nit)/ :-)

spec/index.html
dlongley: Yes to this.

gkellogg: I can't apply both of these, so I'll merge this change into the previous suggestion.

gkellogg (approved):

PR #63, opened 2022-12-22
Markup for logging, including explanations

gkellogg:
... using ~~existing and future~~ examples.

This includes examples from two test cases, ~~which haven't yet been added to the spec~~.

* double circle from [test022](https://w3c.github.io/rdf-canon/tests/#manifest-urdna2015#test022)
* duplicate paths from [test035](https://w3c.github.io/rdf-canon/tests/#manifest-urdna2015#test035)

Note that I added entries to the "How to Read This Document" section, which likely won't show up in the preview.

<dl>
  <dt><span class="details"><span class="summary">Explanation</span></span></dt><dd>
    An expandable area to find a more details non-normative explanation of a
    particular algorithmic step.
    <details>
      <summary>Explanation</summary>
      <p>This area would provide more information about the step involved.</p>
    </details>
  </dd>
  <dt><span class="details"><span class="summary">Logging</span></span></dt><dd>
    An expandable area to find suggestions for implementations to log
    information about processing,
    which may be useful in comparing with other implementations,
    or with logs provided with each test case.
    <details>
      <summary>Logging</summary>
      <p>For example, the following output snippet might
        describe the operation of an implementation using the [[?YAML]] format.</p>
      <pre class="logging yaml" data-transform="updateExample">
        
      </pre>
    </details>
  </dd>
</dl>

Fixes #42.

***

### :boom: Error: 503 Service Unavailable :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 14, 2023, 10:57 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Frdf-canon%2Fb8bc0ab8b6d8f205008f74ed9656b232543eb902%2Fspec%2Findex.html%3FisPreview%3Dtrue)

```
<html><body><h1>503 Service Unavailable</h1>
No server is available to handle this request.
</body></html>

```

[...]

PR #64, opened 2022-12-28
Add example for duplicate-paths test showing rendered log results.

gkellogg:
I added a render of the result of running the duplicate paths test to see how this might look. It results in fairly deep nesting, and I'm not sure it's too useful for the average reader, but might help others.

```turtle
BASE <http://example.com/>
PREFIX : <#>

_:e0 :p1 _:e1 .
_:e1 :p2 "Foo" .
_:e2 :p1 _:e3 .
_:e3 :p2 "Foo" .
```

As much as this does, there are more complicated examples that use more of the algorithm (e.g., test022).

Note that this is an update to PR #63. I'm not sure if it's useful enough to include.

Because this uses data-include, it won't actually show up in PR preview, so you can see it using GitHack [here](https://raw.githack.com/w3c/rdf-canon/duplicated-paths-example/spec/index.html).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/64.html" title="Last updated on Jan 12, 2023, 1:32 AM UTC (20b6bb9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/64/84db619...20b6bb9.html" title="Last updated on Jan 12, 2023, 1:32 AM UTC (20b6bb9)">Diff</a>


gkellogg:
Here's another output for [test022](https://raw.githack.com/ruby-rdf/rdf-normalize/develop/examples/test022.html).

dlongley (approved):
I noticed that the HTML rendering of the logging uses ellipsis when the value is too long sometimes ... we'll want to have a way to expand those for viewing (if we already have a way, I didn't see it).

gkellogg:
The long strings using an ellipsis are tucked in `abbr` elements, so the full value can be revealed by hovering. Perhaps there’s some active interrogation of the View area that could make this dynamic, but it’s beyond me at this point.

Even with the abbreviations, it still makes the view window pretty wide and I’m not really convinced of the value of this kind of example/trace.

More valuable for developers might be to show a YAML and HTML trace for every test, and not try to directly include in the spec.

iherman:
> The long strings using an ellipsis are tucked in `abbr` elements, so the full value can be revealed by hovering. Perhaps there’s some active interrogation of the View area that could make this dynamic, but it’s beyond me at this point.
> 
> Even with the abbreviations, it still makes the view window pretty wide and I’m not really convinced of the value of this kind of example/trace.

Much as it is logical to nest the entries to follow the cycles, the resulting wide table is difficult to follow. I wonder whether it is not better to use this example the same way as example 7, for example, ie, some sort of verbose description of what is happening. That would help the reader to understand the essence of the algorithm without all the details of the YAML output.

> 
> More valuable for developers might be to show a YAML and HTML trace for every test, and not try to directly include in the spec.

Agreed but, actually, I am not even sure that the HTML trace is all that useful for developers. A YAML output might be enough.

gkellogg:
> > The long strings using an ellipsis are tucked in `abbr` elements, so the full value can be revealed by hovering. Perhaps there’s some active interrogation of the View area that could make this dynamic, but it’s beyond me at this point.
> > Even with the abbreviations, it still makes the view window pretty wide and I’m not really convinced of the value of this kind of example/trace.
> 
> Much as it is logical to nest the entries to follow the cycles, the resulting wide table is difficult to follow. I wonder whether it is not better to use this example the same way as example 7, for example, ie, some sort of verbose description of what is happening. That would help the reader to understand the essence of the algorithm without all the details of the YAML output.

I was trying to get away from having to hand author each example, and instead derive something from the logs. In any case, recursion is going to be part of it, so some form of indentation is inevitable. It wasn't immediately apparent how I could interpret the YAML log into something that is less indented, but some areas of it can certainly be customized.

To illustrate all parts of the algorithm will require exploring more examples, and creating these by hand is hard work.

gkellogg:
Although people may take issue with this example, I think it's worth merging into #63.

PR #75, opened 2023-02-01
Rename `_:b*` to `_:e*` in the test 57 (seems to be forgotten in #56)

yamdan:
This is merely additional patching related to #56, where test 057 has remained with its old `_:b*` labels.


iherman:
I believe that was actually intentional. The idea was to have a test does use `_:b` to prove that the implementation doesn't get confused by those and the nodes ids provided by the algorithm.

yamdan:
Thanks. Then, I should have left `_:g` as it was, whereas I think the subject `_:b1` should still be replaced with `_:e0`.

dlongley:
Yeah, we want to have one or more tests with inputs that start with blank node labels that are used internally in the algorithm (i.e., that start with `b`) to ensure an implementation can handle that properly.

gkellogg (changes requested):
Agreed, we shouldn't merge this PR, as that naming is intentional.

yamdan:
I thought such an intention (test with `_:b*`) was already expressed as test063 instead of this test057. Nevertheless, having more than one such test case is fine, so that I will withdraw this PR.

PR #94, opened 2023-04-06
concat separator question; better punc & language

TallTed:
Before merge, this PR should be tweaked to identify the concatenation separator for N-Quads serialization.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/94.html" title="Last updated on Apr 10, 2023, 7:13 PM UTC (cbd681e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/94/64744e0...cbd681e.html" title="Last updated on Apr 10, 2023, 7:13 PM UTC (cbd681e)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2638,14 +2638,15 @@
     created by representing each <a>quad</a> from the <a>normalized dataset</a>
     in <a>canonical n-quads form</a>,
     sorting them into <a>code point order</a>,
-    and concatenating them.
+    and concatenating them with TODO separators.
+    <p class="ednote">TODO separators</p>
     The resulting document has a media type of `application/n-quads`,
     as described in <a data-cite="N-QUADS#sec-mediaReg">C. N-Quads Internet Media Type, File Extension and Macintosh File Type</a>
     of [[N-QUADS]].</p>
 
-  <p>When serializing quads in <a>canonical n-quads form</a>
+  <p>When serializing quads in <a>canonical n-quads form</a>,
     components which are <a>blank nodes</a> MUST be serialized using the
-    canonical label associated with the <a>blank node</a> in the <a>normalized dataset</a>.</p>
+    canonical label associated with each <a>blank node</a> in the <a>normalized dataset</a>.</p>
 
   <aside id="ex-ser-unique-hashes"
         class="example"
--- a/spec/index.html
+++ b/spec/index.html
@@ -2638,7 +2638,7 @@
     created by representing each <a>quad</a> from the <a>normalized dataset</a>
     in <a>canonical n-quads form</a>,
     sorting them into <a>code point order</a>,
-    and concatenating them. (Note that each canonical n-quad ends with a new line, 
+    and concatenating them. (Note that each canonical N-Quads statement ends with a new line, 
     so no additional separators are needed in the concatenation.)
     The resulting document has a media type of `application/n-quads`,
     as described in <a data-cite="N-QUADS#sec-mediaReg">C. N-Quads Internet Media Type, File Extension and Macintosh File Type</a>
--- a/spec/index.html
+++ b/spec/index.html
@@ -2670,7 +2670,7 @@
   </aside>
 </section>
 
-<section id="privacy-considerations">
+<section id="privacy-considerations" class="informative">
   <h2>Privacy Considerations</h2>
 
   <section>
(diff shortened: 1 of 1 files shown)


gkellogg (commented):
There are no concatenation separators. Concatenation does not imply that records are separated by anything; that is often referred to as "joining", where a separator such as a space or newline may be introduced. From [Wikipedia](https://en.wikipedia.org/wiki/Concatenation):

> In [formal language theory](https://en.wikipedia.org/wiki/Formal_language) and [computer programming](https://en.wikipedia.org/wiki/Computer_programming), string concatenation is the operation of joining [character strings](https://en.wikipedia.org/wiki/Character_string_(computer_science)) [end-to-end](https://en.wiktionary.org/wiki/end-to-end). For example, the concatenation of "snow" and "ball" is "snowball". In certain formalisations of [concatenation theory](https://en.wikipedia.org/wiki/Concatenation_theory), also called string theory, string concatenation is a [primitive notion](https://en.wikipedia.org/wiki/Primitive_notion).

iherman:
> There are no concatenation separators. Concatenation does not imply that records are separated by anything; that is often referred to as "joining", where a separator such as a space or newline may be introduced. From [Wikipedia](https://en.wikipedia.org/wiki/Concatenation):
> 
> > In [formal language theory](https://en.wikipedia.org/wiki/Formal_language) and [computer programming](https://en.wikipedia.org/wiki/Computer_programming), string concatenation is the operation of joining [character strings](https://en.wikipedia.org/wiki/Character_string_(computer_science)) [end-to-end](https://en.wiktionary.org/wiki/end-to-end). For example, the concatenation of "snow" and "ball" is "snowball". In certain formalisations of [concatenation theory](https://en.wikipedia.org/wiki/Concatenation_theory), also called string theory, string concatenation is a [primitive notion](https://en.wikipedia.org/wiki/Primitive_notion).

I believe the source of confusion may be that the (canonical) nquad format already include a new line, see https://www.w3.org/TR/n-quads/#sec-intro (see also https://www.w3.org/TR/rdf-canon/#terms-reference). There is no reason to even think adding an extra separator character.

TallTed:
@iherman @gkellogg — 

I had indeed overlooked that the canonical n-quad format ends each quad with a new line. It's surprising that this is only stated explicitly in **§ 1. Introduction**, which is surprisingly normative, though this normativity is only made clear later in **§ 3. Conformance**.

I think that the terminating new line on each canonical n-quad needs to be made more explicit, and something like my adjusted text will help.

spec/index.html
```diff
@@ -2638,14 +2638,15 @@ <h2>Serialization</h2>
     created by representing each <a>quad</a> from the <a>normalized dataset</a>
     in <a>canonical n-quads form</a>,
     sorting them into <a>code point order</a>,
-    and concatenating them.
+    and concatenating them with TODO separators.
+    <p class="ednote">TODO separators</p>
```
TallTed, line 2642:
```suggestion
    and concatenating them. (Note that each canonical n-quad ends with a new line, 
    so no additional separators are needed in the concatenation.)
```

gkellogg:
See https://w3c.github.io/rdf-n-quads/spec/#canonical-quads which is explicit about the required EOL character at the end of each line. Once that doc reaches FPWD, it can be cited.

gkellogg:
The introduction should not be normative; many specs seem to not mark their introductions or abstracts as informative. But, we should mark our introductions as informative (and the RDF-star specs too).

gkellogg (approved):

dlongley (approved):

spec/index.html
```diff
@@ -2638,14 +2638,15 @@ <h2>Serialization</h2>
     created by representing each <a>quad</a> from the <a>normalized dataset</a>
     in <a>canonical n-quads form</a>,
     sorting them into <a>code point order</a>,
-    and concatenating them.
+    and concatenating them. (Note that each canonical n-quad ends with a new line,
```
yamdan, line 2641:
```suggestion
    and concatenating them. (Note that each canonical N-Quads statement ends with a new line, 
```
If I am not mistaken, the term 'n-quad' may not be well-defined. It seems a bit more appropriate to use 'statement' instead.

yamdan (commented):
Thank you @TallTed for the clarification. I have added one small suggestion below.

spec/index.html
gkellogg: The N-Quads grammar defines a [statement](https://w3c.github.io/rdf-n-quads/spec/#grammar-production-statement) which could be referenced. Not a term, but a grammar rule.

PR #97, opened 2023-05-02
Update algorithm output to be the serialized canonical form

### LINKED ISSUE #4: What is the output of the c14n algorithm?
Following the discussion that just happened at the TPAC joint meeting with the VC.

An option is to return one particular serialization.
An another option is to return a mapping from bnodes to labels.

I prefer the second solution as it is more generic. It does not preclude which serialization to use downstream. It actually does not impose that you serialize the dataset (imagine storing a dataset in a triple store with canonical blank node labels).

### LINKED ISSUE #92: Figure out RDH process flow and additional outputs
Following the TPAC joint meeting discussion and subsequent issue w3c/rch-rdc#4 raised by @pchampin, I think it could be a good entry point to determine the whole possible process flow from an RDF document to an RDF hash value. Furthermore, we could consider additional flows to get the other types of outputs besides a simple hash value, which is helpful for ZKP-friendly multi-message signatures such as [BBS+](https://datatracker.ietf.org/doc/draft-looker-cfrg-bbs-signatures/).

The blue-colored parts in the following figure show my current understanding of calculating a hash value from an RDF document:
20220927_rch
We start with an **RDF document**, such as a JSON-LD document describing that Alice knows Bobby.
We can **deserialize** the document to obtain an abstract **RDF dataset**, an unordered set of RDF graphs/triples.
Using **RDF canonicalization algorithms** (to be standardized in this WG), the dataset can be canonicalized into a **canonicalized RDF dataset**, where all the blank nodes are deterministically labeled.
Here I assume that the canonicalized RDF dataset is an abstract dataset aligned with the WG charter, related to Issue #4.
We can **serialize** the canonicalized RDF dataset, with some whitespace normalizations, into an N-Quads document, which I temporarily name as a **semi-canonicalized RDF document**, where the order of RDF quads is not determined yet.
Then, we can lexicographically **sort** it to get a **canonicalized RDF document**, where all the blank node identifiers and the order of quads are deterministically generated from the input RDF document.
The canonicalized RDF document can be **hashed** using an appropriate hash function to obtain a **canonicalized RDF hash value**.
The above flow describes my current understanding of a simple RDF hashing, where we could revise some terminologies later.

Besides the blue-colored main flow, we can also discuss other valuable branches, especially for multi-message signature schemes like BBS+.
For example, [LDP-BBS2020](https://w3c-ccg.github.io/ldp-bbs2020/) splits the canonicalized RDF document into statements, regarded as messages to be signed by the BBS+ signature scheme. This statement-wise splitting with BBS+ allows us to do statement-wise selective disclosure.
Furthermore, a term-wise encoding, used in our modified version of LDP-BBS2020 ([paper](https://sako-lab.jp/download.php?article=ssr2022_proceedings_dan.pdf) / [demo](https://zkp-ld.org/)) mentioned in TPAC meeting by @OR13, splits the canonicalized RDF document into RDF **terms** to enable term-wise selective disclosure and predicate proofs with BBS+.
I wonder if these canonicalized RDF statements/terms could also be worth considering as the extended outputs of the RDF "hash" in this WG.

gkellogg:
Update definition of "normalized dataset" to be an RDF dataset and an ordered map mapping canonical labels to blank nodes within that dataset.

Update the result of the canonicalization algorithm to be the serialized canonical form of that normalized dataset.

Fixes #4. Fixes #92.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/97.html" title="Last updated on May 5, 2023, 4:59 PM UTC (fa8ecd5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/97/d69576c...fa8ecd5.html" title="Last updated on May 5, 2023, 4:59 PM UTC (fa8ecd5)">Diff</a>


peacekeeper:
I think to complete #4, this is still missing some language about how implementations may change stable blank node identifiers before serialization, or change the order of serialized N-Quads, or something like that.

spec/index.html
```diff
@@ -409,16 +411,19 @@ <h3>Terms defined by cited specifications</h3>
...
+    The problem requires directed
     graphs to be deterministically ordered into sets of nodes and edges. This
     is easy to do when all of the nodes have globally-unique identifiers, but
     can be difficult to do when some of the nodes do not. Any nodes without
     globally-unique identifiers must be issued deterministic identifiers.</p>
 
-  <p class="ednote">Strictly speaking, the normalized dataset must be serialized to be stable, as within a dataset, blank node identifiers have no meaning. This specification defines a <a>normalized dataset</a> to include stable identifiers for blank nodes, but practical uses of this will always generate a canonical serialization of such a dataset.</p>
+  <p class="note">
+    This specification defines a <a>normalized dataset</a> to include stable identifiers for blank nodes,
+    but practical uses of this will always generate a canonical serialization of such a dataset.</p>
```
TallTed, line 426:
```suggestion
    This specification defines a <a>normalized dataset</a> to include stable identifiers for blank nodes,
    practical uses of which will always generate a canonical serialization of such a dataset.</p>
```

TallTed (changes requested):
(editorial) Fairly minor but clarifying, I think.

gkellogg:
> I think to complete #4, this is still missing some language about how implementations may change stable blank node identifiers before serialization, or change the order of serialized N-Quads, or something like that.

I thought about that, but not sure where the wording would go. The branching point is at step 7 in the Canonicalization Algorithm. Clearly, the serializer could be called with an alternative dataset that uses a different mapping and an implementation could insert steps before step 7 to perform some operations on the normalized dataset to create such an alternative map.

spec/index.html
```diff
@@ -297,10 +297,12 @@ <h3>Terms defined by this specification</h3>
...
       <dt><dfn>normalized dataset</dfn></dt>
       <dd>The immutable, abstract <a>RDF dataset</a> and a set of normalized
-        <a>blank node identifiers</a> that are produced as output by the algorithm.
-        A <a>normalized dataset</a> is a restriction on an <a>RDF dataset</a>
-        where all nodes are labeled, and <a>blank nodes</a> are labeled with canonical <a>blank node identifiers</a>
-        consistent with running this algorithm on a base <a>RDF dataset</a>.
+        <a>blank node identifiers</a> that are produced as consistent with this algorithm.
+        A <a>normalized dataset</a> is the combination of an <a>RDF dataset</a>
+        and an <a>ordered map</a> where [=map/keys=]
+        are canonical <a>blank node identifiers</a> generated by this
```
gkellogg, line 303: The choice to have keys be identifiers and values be blank nodes may be backwards, but it makes sense when considering that it is ordered. Alternatively, the key/value could be swapped, and it could be described as being ordered on values, rather than keys. In any case, it's a 1:1 mapping, so it can be reversed easily enough in code.

spec/index.html
```diff
@@ -297,10 +297,11 @@ <h3>Terms defined by this specification</h3>
         the algorithm.</dd>
       <dt><dfn>normalized dataset</dfn></dt>
       <dd>The immutable, abstract <a>RDF dataset</a> and a set of normalized
-        <a>blank node identifiers</a> that are produced as output by the algorithm.
-        A <a>normalized dataset</a> is a restriction on an <a>RDF dataset</a>
-        where all nodes are labeled, and <a>blank nodes</a> are labeled with canonical <a>blank node identifiers</a>
-        consistent with running this algorithm on a base <a>RDF dataset</a>.
+        <a>blank node identifiers</a> that are produced as consistent with this algorithm.
```
iherman, line 300: Is the first sentence in this paragraph still necessary? The second sentence, ie, the one starting with

```html
A <a>normalized dataset</a> is the combination…
```

says the same thing in a more precise format. I would consider removing the first sentence altogether.

iherman (approved):

Issue #89, opened 2023-04-03
Support ordered input dataset or "list of quads" and optional mapping from input indices to output indices


dlongley:
Starting with my comment here: https://github.com/w3c/rdf-canon/issues/86#issuecomment-1494529549

Some discussion spawned around the need, as an optional output, a mapping of quad input indices to quad output indices for selective disclosure use cases.

@gkellogg made [this comment](https://github.com/w3c/rdf-canon/issues/86#issuecomment-1494911219):

> Just be be clear on what we're talking about, the [input dataset](https://w3c.github.io/rdf-canon/spec/#dfn-input-dataset) is unordered and no blank node labels are persistent or possibly even present.

Which implies that we might want to also take an ordered list of quads as an optional alternative input to the algorithm. Or perhaps we can describe the RDF abstract dataset as being optionally represented as such -- for the case where this mapping output is desirable. Notably, the presence (or lack thereof) of input blank node labels in this case is not relevant.

dlongley:
To elaborate on the problem a bit more here:

The goal is to support selective disclosure use cases with verifiable credentials.

The main scenario is this:

A holder of a verifiable credential must be able to reveal a subset of the quads from the verifiable credential to a verifier. The verifier must be able to reproduce the same bnode labels that were used when all of the quads from the verifiable credential were canonized and cryptographically signed by the issuer.

The problem here lies in the fact that canonizing a subset of a dataset can produce different bnode labels (for the same blank nodes) from canonizing the entire dataset.

The process the holder goes through to prepare to selectively disclose a verifiable credential to a verifier looks something like this:

1. Canonize the VC's dataset to get canonical bnode labels.
2. Optionally, modify those bnode labels in a pseudo-random way.
3. Pick out just the quads they want to disclose.
4. Knowing the verifier will only see those quads, canonize only the selected quads to get the bnode labels the verifier will see.
5. *Somehow* produce a map of the bnode labels the verifier will see to the labels used after step 2.
6. Share that mapping with the verifier when selectively disclosing the verifiable credential.

Step 5 is the step we need to enable for the holder.

Once we've enabled step 5 for the holder, then the verifier can do this:

1. Canonize the selectively disclosed VC's dataset to get bnode labels that need to be mapped.
2. Use the mapping from the holder to transform the bnode labels.
3. Cryptographically verify the transformed data.

yamdan:
@dlongley 
[...]

PR #99, opened 2023-05-12
Add privacy considerations

### LINKED ISSUE #84: Some privacy considerations
As a starting point for preparing Privacy Considerations (#70), I am trying to organize some existing discussions about privacy aspects of the URDNA2015, including:
- [Dave's comment in #4](https://github.com/w3c/rdf-canon/issues/4#issuecomment-1262602679)
- [Blank node labels may leak information · Issue #60 · w3c-ccg/ldp-bbs2020](https://github.com/w3c-ccg/ldp-bbs2020/issues/60)
- Academic articles by Vasilis Kalos [(1)](https://mm.aueb.gr/master_theses/polyzos/2021-Kalos.pdf), [(2)](https://mm.aueb.gr/publications/fd17e44e-9b7a-477d-b948-81a71db53b06.pdf)

While the discussions here mainly focus on URDNA2015, I believe we can apply similar arguments to the other RDF canonicalization algorithms.

-----

Privacy considerations here are primarily worth discussing when the canonicalization scheme is used for privacy-respecting signed RDF dataset. 

The following issues are worth discussing when the canonicalization scheme is used for privacy-respecting signed RDF datasets and are likely acceptable for other use cases. One of the former examples is a verifiable credential with selective disclosure.

Selective disclosure is the ability for someone to share only some of the statements from a signed dataset, without harming the ability of the recipient to verify the authenticity of those selected statements. (Note: copied from [Dave's comment in #4](https://github.com/w3c/rdf-canon/issues/4#issuecomment-1262602679))

The normalized dataset, the output of the canonicalization algorithm described in this specification, may leak partial information about undisclosed statements and help the adversary correlate the original and disclosed datasets.

## a) possible leakage via canonical labeling

If a dataset contains at least two blank nodes, the canonical labeling can be exploited to guess the undisclosed quad in the dataset.

For example, let us assume we have the following dataset to be signed. (Note: this person is fictitious, prepared only to make this example work.)

```turtle
# original dataset
_:b0 <http://schema.org/address> _:b1 .
_:b0 <http://schema.org/familyName> "Jarrett" .
_:b0 <http://schema.org/gender> "Female" .  # gender === Female
_:b0 <http://schema.org/givenName> "Ali" .
_:b0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
_:b1 <http://schema.org/addressCountry> "United States" .
_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/PostalAddress> .
```

Using URDNA2015, we can obtain the normalized dataset with canonical labels  sorted in the canonical (code-point) order.

```turtle
# normalized dataset
_:c14n0 <http://schema.org/addressCountry> "United States" .
_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/PostalAddress> .
_:c14n1 <http://schema.org/address> _:c14n0 .
_:c14n1 <http://schema.org/familyName> "Jarrett" .
_:c14n1 <http://schema.org/gender> "Female" .  # gender === Female
_:c14n1 <http://schema.org/givenName> "Ali" .
_:c14n1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
```

The signer (or issuer) can generate a signature for the dataset by first hashing each statement and then signing them using a multi-message digital signature scheme like BBS+. The resulting dataset with signature is passed to the holder, who can control whether or not to disclose each statement while maintaining their verifiability.

Let us say that the holder wants to show her attributes except for `gender` to a verifier. Then the holder should disclose the following partial dataset. (Note: proofs omitted here for brevity)

```turtle
# disclosed dataset
_:c14n0 <http://schema.org/addressCountry> "United States" .
_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/PostalAddress> .
_:c14n1 <http://schema.org/address> _:c14n0 .
_:c14n1 <http://schema.org/familyName> "Jarrett" .
########### 5th statement is unrevealed ##########
_:c14n1 <http://schema.org/givenName> "Ali" .
_:c14n1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
```

However, in this example, anyone can guess the unrevealed statement by exploiting the canonical labels and order.

Since the dataset was sorted in the canonical order, we can get to know that the hidden statement must start with `_:c14n1 <http://schema.org/[f-g]`, which helps us guess that the hidden predicate is `<http://schema.org/gender>` with high probability.
Alternatively, we can assume that the guesser already has such knowledge via the public credential schema.

Then, if the canonical labeling produces different results depending on the gender value, we can use it to deduce the gender value.
In fact, this example produces different results depending on whether the gender is `Female` or `Male`.
(Note: I ignored the other types of gender just for brevity)

The following example shows that "gender = Male" yields different canonical labeling.

```turtle
# hypothetical normalized dataset
_:c14n0 <http://schema.org/address> _:c14n1 .
_:c14n0 <http://schema.org/familyName> "Jarrett" .
_:c14n0 <http://schema.org/gender> "Male" .  # gender === Male
_:c14n0 <http://schema.org/givenName> "Ali" .
_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
_:c14n1 <http://schema.org/addressCountry> "United States" .
_:c14n1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/PostalAddress> .
```

So the verifier should have obtained the following dataset if `gender` had the value `Male`, which differs from the revealed dataset, so the verifier can conclude that the `gender` is `Female`.

```turtle
# hypothetical disclosed dataset
_:c14n0 <http://schema.org/address> _:c14n1 .
_:c14n0 <http://schema.org/familyName> "Jarrett" .
########### 3rd statement is unrevealed ##########
_:c14n0 <http://schema.org/givenName> "Ali" .
_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
_:c14n1 <http://schema.org/addressCountry> "United States" .
_:c14n1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/PostalAddress> .
```

Note that we can use the same approach to guess non-boolean values if the range of possible values is still a reasonable (small) size for a guesser to try all the possibilities.

By making the canonicalization process private, we can prevent a brute-forcing attacker from trying to see the labeling change by trying multiple possible attribute values.
For example, we can use a HMAC instead of a hash function in the canonicalization algorithm. Alternatively, we can add a secret random nonce (always undisclosed) into the dataset.
Note that these workarounds force dataset issuers and holders to manage shared secrets securely.
We also note that these workarounds adversely affect the unlinkability described below because canonical labeling now varies depending on the secret shared by the issuer and the holder, which will help correlate them.

## b) possible leakage via canonical sorting

The canonical order can leak unrevealed information even without canonical labelings.

Let us assume that the holder has the following signed dataset, sorted in the canonical (code-point) order.

```turtle
:a <http://schema.org/children> "Albert" .
:a <http://schema.org/children> "Alice" .
:a <http://schema.org/children> "Allie" .
:a <http://schema.org/name> "John Smith" .
```

If the holder wants to hide the statement for their second child for any reason, the disclosed dataset now looks like this:

```turtle
:a <http://schema.org/children> "Albert" .
########### 2nd statement is unrevealed ##########
:a <http://schema.org/children> "Allie" .
:a <http://schema.org/name> "John Smith" .
```

Knowing that these statements are sorted in the canonical order, we can guess that the hidden statement must start with `:a <http://schema.org/children> "Al`, which leaks the subject (`:a`), predicate (`<http://schema.org/children>`) and the first two letters of the object (`"Al"`) in the hidden statement.

To avoid this leakage, the dataset issuer can randomly shuffle the normalized statements before signing and issuing them to the holder, preventing others from guessing undisclosed information from the canonical order.
However, similar to the workarounds mentioned above, this workaround also adversely affects unlinkability. This is because there are $n!$ different permutations for shuffling $n$ statements, and whichever one is used will help correlate the dataset.

## c) possible linking via canonical labeling

Unlinkability ensures that no correlatable data are used in a signed dataset while still providing some level of trust, the sufficiency of which must be determined by each verifier. (Note: based on the description in the VC Data Integrity spec)

While canonical sorting works better for unlinkability, canonical labeling can be exploited to break it.
The total number of canonical labelings for a dataset with $n$ blank nodes is $n!$, which is not controllable by the issuer.
It means that the herd constructed as a result of selective disclosure will be split into $n!$ pieces due to the canonical labeling, which reduces unlinkability.

For example, let us assume that an employee of the small company "example.com" shows its employee ID dataset without their name like this:

```turtle
# disclosed dataset
########### 1st statement is unrevealed ##########
_:c14n0 <http://schema.org/worksFor> _:c14n1 .
_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
_:c14n1 <http://schema.org/address> _:c14n2 .
_:c14n1 <http://schema.org/geo> _:c14n3 .
_:c14n1 <http://schema.org/name> "example.com" .
_:c14n2 <http://schema.org/addressCountry> "United States" .
_:c14n3 <http://schema.org/latitude> "0.0" .
_:c14n3 <http://schema.org/longitude> "0.0" .
```

The verifier can always trace this person without knowing their name if this company has only three employees with the following employee ID datasets. 

```turtle
# normalized dataset 1
_:c14n0 <http://schema.org/address> _:c14n1 .
_:c14n0 <http://schema.org/geo> _:c14n3 .
_:c14n0 <http://schema.org/name> "example.com" .
_:c14n1 <http://schema.org/addressCountry> "United States" .
_:c14n2 <http://schema.org/name> "Jayden Doe" .
_:c14n2 <http://schema.org/worksFor> _:c14n0 .
_:c14n2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
_:c14n3 <http://schema.org/latitude> "0.0" .
_:c14n3 <http://schema.org/longitude> "0.0" .
```

```turtle
# normalized dataset 2
_:c14n0 <http://schema.org/address> _:c14n1 .
_:c14n0 <http://schema.org/geo> _:c14n2 .
_:c14n0 <http://schema.org/name> "example.com" .
_:c14n1 <http://schema.org/addressCountry> "United States" .
_:c14n2 <http://schema.org/latitude> "0.0" .
_:c14n2 <http://schema.org/longitude> "0.0" .
_:c14n3 <http://schema.org/name> "Morgan Doe" .
_:c14n3 <http://schema.org/worksFor> _:c14n0 .
_:c14n3 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
```

```turtle
# normalized dataset 3
_:c14n0 <http://schema.org/name> "Johnny Smith" .
_:c14n0 <http://schema.org/worksFor> _:c14n1 .
_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
_:c14n1 <http://schema.org/address> _:c14n2 .
_:c14n1 <http://schema.org/geo> _:c14n3 .
_:c14n1 <http://schema.org/name> "example.com" .
_:c14n2 <http://schema.org/addressCountry> "United States" .
_:c14n3 <http://schema.org/latitude> "0.0" .
_:c14n3 <http://schema.org/longitude> "0.0" .
```

The canonicalization in this example produces different labelings for these three employees, which helps anyone to correlate their activities even if they do not reveal their names in the dataset.

By determining some "template" for each anonymous set (or herd) and fixing the canonical labeling and canonical order used in the anonymous set, we can achieve a certain unlinkability.
Alternatively, we might be able to generate a kind of "on-the-fly" template proposed in [Dave's comment in #4](https://github.com/w3c/rdf-canon/issues/4#issuecomment-1262602679), which seems ineffective in some cases (see #8).

yamdan:
This PR is a draft of Privacy Considerations section, which consists of (1) data leakage in selective disclosure schemes and (2) unlinkability, based on the discussion in #84 . While we could additionally mention a kind of **tradeoff** between these two notions, I have not dealt with it because I am still not fully convinced (confident) that such a tradeoff does indeed exist...

Fixes #84.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/99.html" title="Last updated on May 19, 2023, 8:12 AM UTC (ae75903)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/99/7bc4143...ae75903.html" title="Last updated on May 19, 2023, 8:12 AM UTC (ae75903)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2675,7 +2675,7 @@
   <h2>Privacy Considerations</h2>
 
   <p>In general, RDF datasets are used for representing and conveying arbitrary information,
-    which may include personally identifiable information which may be used for correlating
+    which may include personally identifiable information that may be used for correlating
     other with data and disclose arbitrary information.
     After canonicalizing a dataset, even if some information is removed,
     it may be possible for a third party to infer omitted data.
@@ -2697,7 +2697,7 @@ disclose.
     <p>Selective disclosure is the ability for someone to share only some of the statements from a signed dataset, without
       harming the ability of the recipient to verify the authenticity of those selected statements.</p>
     
-    <p>The output of the canonicalization algorithm described in this specification, may leak partial
+    <p>The output of the canonicalization algorithm, described in this specification, may leak partial
       information about undisclosed statements and help the adversary correlate the original and disclosed datasets.</p>
     
     <section id="privacy-considerations-leakage-labeling">
@@ -2741,7 +2741,7 @@ disclose.
         holder, who can control whether or not to disclose each statement while maintaining their verifiability.</p>
       
       <p>Let us say that the holder wants to show their attributes except for `gender` to a verifier. Then the holder should
-        disclose the following partial dataset. (Note: proofs omitted here for brevity)</p>
+        disclose the following partial dataset. </p>
 
       <pre id="ex-pc-leakage-labeling-disclosed-dataset" class="example" data-transform="updateExample" title="Disclosed Dataset">
         <!--          
@@ -2758,13 +2758,13 @@ disclose.
       <p>However, in this example, anyone can guess the unrevealed statement by exploiting the canonical identifiers and order.</p>
       
       <p>Since the dataset was sorted in the canonical order, we can get to know that the hidden statement must start with
-        `_:c14n1 &lt;http://schema.org/[f-g]`, which helps us guess that the hidden predicate is
-        `&lt;http://schema.org/gender&gt;` with high probability. Alternatively, we can assume that the guesser already has
+        `_:c14n1 &lt;http://schema.org/[f-g]`, which helps us guess that the hidden predicate is, with high probability,
+        `&lt;http://schema.org/gender&gt;` . Alternatively, we can assume that the guesser already has
         such knowledge via the public credential schema.</p>
       
       <p>Then, if the canonical labeling produces different results depending on the gender value, we can use it to deduce the
-        gender value. In fact, this example produces different results depending on whether the gender is `Female` or `Male`.
-        (Note: ignored the other types of gender just for brevity)</p>
+        gender value. Indeed, this example produces different results depending on whether the gender is `Female` or `Male`.
+        (Note: ignored the other types of gender for the sake of the example.)</p>
       
       <p>The following example shows that `gender` = `Male` yields different canonical labeling.</p>
 
@@ -2795,8 +2795,8 @@ disclose.
         -->
       </pre>
 
-      <p>Note that we can use the same approach to guess non-boolean values if the range of possible values is still a
-        reasonable (small) size for a guesser to try all the possibilities.</p>
+      <p>Note that we can use the same approach to guess non-boolean values if the range of possible values is still of a
+        reasonably small size for to try all the possibilities.</p>
 
       <p>By making the canonicalization process private, we can prevent a brute-forcing attacker from trying to see the
         labeling change by trying multiple possible attribute values.
@@ -2804,7 +2804,7 @@ disclose.
         a secret random nonce (always undisclosed) into the dataset.
         Note that these workarounds force dataset issuers and holders to manage shared secrets securely.
         We also note that these workarounds adversely affect the unlinkability described below because canonical labeling now
-        varies depending on the secret shared by the issuer and the holder, which will help correlate them.</p>
+        varies depending on the secret shared by the issuer and the holder, which may help to correlate them.</p>
     </section>
     
     <section id="privacy-considerations-leakage-sorting">
@@ -2842,7 +2842,7 @@ disclose.
       <p>To avoid this leakage, the dataset issuer can randomly shuffle the normalized statements before signing and issuing
         them to the holder, preventing others from guessing undisclosed information from the canonical order.
         However, similar to the workarounds mentioned above, this workaround also adversely affects unlinkability. This is
-        because there are $n!$ different permutations for shuffling $n$ statements, and whichever one is used will help
+        because there are `n!` different permutations for shuffling `n` statements, and whichever one is used will help
         correlate the dataset.</p>
     </section>
   </section>
@@ -2854,9 +2854,9 @@ disclose.
       trust, the sufficiency of which must be determined by each verifier. </p>
 
     <p>While canonical sorting works better for unlinkability, canonical labeling can be exploited to break it.
-      The total number of canonical labelings for a dataset with $n$ blank nodes is $n!$, which is not controllable by the
+      The total number of canonical labelings for a dataset with `n` blank nodes is `n!`, which is not controllable by the
       issuer.
-      It means that the herd constructed as a result of selective disclosure will be split into $n!$ pieces due to the
+      It means that the herd constructed as a result of selective disclosure will be split into `n!` pieces due to the
       canonical labeling, which reduces unlinkability.</p>
 
     <p>For example, let us assume that an employee of the small company "example.com" shows its employee ID dataset without
@@ -2931,7 +2931,7 @@ disclose.
       correlate their activities even if they do not reveal their names in the dataset.</p>
 
     <p>By determining some "template" for each anonymous set (or herd) and fixing the canonical labeling and canonical order
-      used in the anonymous set, we can achieve a certain unlinkability.</p>
+      used in the anonymous set, we can achieve a certain level of unlinkability.</p>
   </section>
 </section>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -2707,7 +2707,7 @@ disclose.
         quad in the dataset.</p>
       
       <p>For example, let us assume we have the following dataset to be signed,
-        describing the fictitious person, "Ali Jarret".</p>
+        describing the fictitious person, "Ali Jarrett".</p>
 
       <pre id="ex-pc-leakage-labeling-original-dataset" class="example" data-transform="updateExample" title="Original Dataset">
         <!--
--- a/spec/index.html
+++ b/spec/index.html
@@ -2674,8 +2674,12 @@
 <section id="privacy-considerations" class="informative">
   <h2>Privacy Considerations</h2>
 
-  <section>
-    <h3>Selective Disclosure Schemes</h3>
+  <p>Privacy considerations here are primarily worth discussing when the canonicalization scheme is used for
+    privacy-respecting signed RDF dataset and are likely acceptable for other use cases. One of the former examples is a
+    verifiable credential with selective disclosure.</p>
+
+  <section id="privacy-considerations-leakage">
+    <h3>Data Leakage in Selective Disclosure Schemes</h3>
 
     <p class="issue" data-number="70" title="Dataset structure might reveal information">
 Add text that warns implementers using this specification in selective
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -2674,8 +2674,12 @@ <h2>Serialization</h2>
 <section id="privacy-considerations" class="informative">
   <h2>Privacy Considerations</h2>
 
-  <section>
-    <h3>Selective Disclosure Schemes</h3>
+  <p>Privacy considerations here are primarily worth discussing when the canonicalization scheme is used for
+    privacy-respecting signed RDF dataset and are likely acceptable for other use cases. One of the former examples is a
+    verifiable credential with selective disclosure.</p>
```
gkellogg, line 2679: This probably needs re-wording. I'll come up with a suggestion at a later point.

spec/index.html
```diff
@@ -2686,8 +2690,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+      information about undisclosed statements and help the adversary correlate the original and disclosed datasets.</p>
+    
+    <section id="privacy-considerations-leakage-labeling">
+      <h4>Possible Leakage via Canonical Labeling</h4>
+    
+      <p>If a dataset contains at least two blank nodes, the canonical labeling can be exploited to guess the undisclosed
+        quad in the dataset.</p>
+      
+      <p>For example, let us assume we have the following dataset to be signed. (Note: this person is fictitious, prepared
+        only to make this example work.)</p>
```
gkellogg, line 2707:
```suggestion
      <p>For example, let us assume we have the following dataset to be signed,
        describing the fictitious woman "Ali Jarret".</p>
```

spec/index.html
```diff
@@ -2686,8 +2690,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+          _:b0 <http://schema.org/gender> "Female" .  # gender === Female
+          _:b0 <http://schema.org/givenName> "Ali" .
+          _:b0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
+          _:b1 <http://schema.org/addressCountry> "United States" .
+          _:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/PostalAddress> .
+        -->
+      </pre>
+
+      <p>Using <a href="#canon-algorithm" class="sectionRef"></a>, we can obtain the <a>serialized canonical form</a> of the
+          <a>normalized dataset</a>, where all the blank nodes are serialized using the canonical labels.</p>
```
gkellogg, line 2722:
```suggestion
          <a>normalized dataset</a>, where all the blank nodes are serialized using the canonical identifiers.</p>

File: spec/index.html [hunk 4]
```diff
@@ -2686,8 +2690,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+          _:c14n1 <http://schema.org/givenName> "Ali" .
+          _:c14n1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
+        -->
+      </pre>
+
+      <p>The signer can generate a signature for the dataset by first hashing each statement and then signing them
+        using a multi-message digital signature scheme like BBS+. The resulting dataset with signature is passed to the
+        holder, who can control whether or not to disclose each statement while maintaining their verifiability.</p>
+      
+      <p>Let us say that the holder wants to show her attributes except for `gender` to a verifier. Then the holder should
```
gkellogg, line 2740: Actually using pronouns such as "his" and "her" ends up getting flagged, so prefer gender neutral terms.

```suggestion
      <p>Let us say that the holder wants to show their attributes except for `gender` to a verifier. Then the holder should
```

spec/index.html
```diff
@@ -2686,8 +2690,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+          _:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/PostalAddress> .
+          _:c14n1 <http://schema.org/address> _:c14n0 .
+          _:c14n1 <http://schema.org/familyName> "Jarrett" .
+          ########### 5th statement is unrevealed ##########
+          _:c14n1 <http://schema.org/givenName> "Ali" .
+          _:c14n1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person> .
+        -->
+      </pre>
+
+      <p>However, in this example, anyone can guess the unrevealed statement by exploiting the canonical labels and order.</p>
```
gkellogg, line 2755:
```suggestion
      <p>However, in this example, anyone can guess the unrevealed statement by exploiting the canonical identifiers and order.</p>
```

gkellogg (commented):
This is a great analysis, and I made a couple of suggestions, but we may evolve the text further. I note that some of the text discusses alternatives (HMAC instead of Hash, re-ordering quads) that don't represent the decisions actually taken by the spec. I'm wondering if we need to provide a rationale, or suggest other ways to mitigate these vulnerabilities compatible with the spec.

spec/index.html
TallTed:
```suggestion
      <p>For example, let us assume we have the following dataset to be signed,
        describing the fictitious person, "Ali Jarret".</p>
```

spec/index.html
gkellogg, line 2679: Here's an idea about how we might frame the introduction to this section.

```suggestion
  <p>In general, RDF datasets are used for representing and conveying arbitrary information,
    which may include personally identifiable information which may be used for correlating
    other with data and disclose arbitrary information.
    After canonicalizing a dataset, even if some information is removed,
    it may be possible for a third party to infer omitted data.
    Applications should consider the potential for exposing personally identifiable information when constructing and conveying datasets.</p>
```

yamdan: Thank you for your re-wording! I applied your suggestion.

spec/index.html
```diff
@@ -2674,8 +2674,15 @@ <h2>Serialization</h2>
 <section id="privacy-considerations" class="informative">
   <h2>Privacy Considerations</h2>
 
-  <section>
-    <h3>Selective Disclosure Schemes</h3>
+  <p>In general, RDF datasets are used for representing and conveying arbitrary information,
+    which may include personally identifiable information which may be used for correlating
```
iherman, line 2678:
```suggestion
    which may include personally identifiable information that may be used for correlating
```

(only stylistic change)

spec/index.html
```diff
@@ -2686,8 +2693,246 @@ <h3>Selective Disclosure Schemes</h3>
 which might be enough to disclose information beyond what the discloser intended to
 disclose.
     </p>
+
+    <p>Selective disclosure is the ability for someone to share only some of the statements from a signed dataset, without
+      harming the ability of the recipient to verify the authenticity of those selected statements.</p>
+    
+    <p>The output of the canonicalization algorithm described in this specification, may leak partial
```
iherman, line 2700:
```suggestion
    <p>The output of the canonicalization algorithm, described in this specification, may leak partial
```

spec/index.html
```diff
@@ -2686,8 +2693,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+
+      <p>However, in this example, anyone can guess the unrevealed statement by exploiting the canonical identifiers and order.</p>
+      
+      <p>Since the dataset was sorted in the canonical order, we can get to know that the hidden statement must start with
+        `_:c14n1 &lt;http://schema.org/[f-g]`, which helps us guess that the hidden predicate is
+        `&lt;http://schema.org/gender&gt;` with high probability. Alternatively, we can assume that the guesser already has
+        such knowledge via the public credential schema.</p>
+      
+      <p>Then, if the canonical labeling produces different results depending on the gender value, we can use it to deduce the
+        gender value. In fact, this example produces different results depending on whether the gender is `Female` or `Male`.
```
iherman, line 2766:
```suggestion
        gender value. Indeed, this example produces different results depending on whether the gender is `Female` or `Male`.
```

spec/index.html
```diff
@@ -2686,8 +2693,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+        (Note: ignored the other types of gender just for brevity)</p>
```
iherman, line 2767:
```suggestion
        (Note: ignored the other types of gender for the sake of the example.)</p>
```

spec/index.html
```diff
@@ -2686,8 +2693,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+        -->
+      </pre>
+
+      <p>Note that we can use the same approach to guess non-boolean values if the range of possible values is still a
+        reasonable (small) size for a guesser to try all the possibilities.</p>
+
+      <p>By making the canonicalization process private, we can prevent a brute-forcing attacker from trying to see the
+        labeling change by trying multiple possible attribute values.
+        For example, we can use a HMAC instead of a hash function in the canonicalization algorithm. Alternatively, we can add
+        a secret random nonce (always undisclosed) into the dataset.
```
iherman, line 2804: This is a similar remark, with the difference that we have not, in my recollection, discussed to add a nonce to the algorithm (I presume to be used with the hash function). I think it would be possible and probably worthwhile to allow that, but that would require a separate WG resolution (and further test cases, b.t.w.)

spec/index.html
```diff
@@ -2686,8 +2693,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+        Note that these workarounds force dataset issuers and holders to manage shared secrets securely.
+        We also note that these workarounds adversely affect the unlinkability described below because canonical labeling now
+        varies depending on the secret shared by the issuer and the holder, which will help correlate them.</p>
```
iherman, line 2807:
```suggestion
        varies depending on the secret shared by the issuer and the holder, which may help to correlate them.</p>
```

spec/index.html
```diff
@@ -2686,8 +2693,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+      </pre>
+
+      <p>Knowing that these statements are sorted in the canonical order, we can guess that the hidden statement must start
+        with `:a &lt;http://schema.org/children&gt; "Al`, which leaks the subject (`:a`), predicate
+        (`&lt;http://schema.org/children&gt;`) and the first two letters of the object (`"Al"`) in the hidden statement.</p>
+      
+      <p>To avoid this leakage, the dataset issuer can randomly shuffle the normalized statements before signing and issuing
+        them to the holder, preventing others from guessing undisclosed information from the canonical order.
+        However, similar to the workarounds mentioned above, this workaround also adversely affects unlinkability. This is
+        because there are $n!$ different permutations for shuffling $n$ statements, and whichever one is used will help
```
iherman, line 2845:
```suggestion
        because there are `n!` different permutations for shuffling `n` statements, and whichever one is used will help
```

spec/index.html
```diff
@@ -2686,8 +2693,246 @@ <h3>Selective Disclosure Schemes</h3>
...
+          _:c14n3 <http://schema.org/longitude> "0.0" .
+        -->
+      </pre>
+    </aside>
+
+    <p>The canonicalization in this example produces different labelings for these three employees, which helps anyone to
+      correlate their activities even if they do not reveal their names in the dataset.</p>
+
+    <p>By determining some "template" for each anonymous set (or herd) and fixing the canonical labeling and canonical order
+      used in the anonymous set, we can achieve a certain unlinkability.</p>
```
iherman, line 2934:
```suggestion
      used in the anonymous set, we can achieve a certain level of unlinkability.</p>
```

iherman (approved):

iherman:
> I note that some of the text discusses alternatives (HMAC instead of Hash, re-ordering quads) that don't represent the decisions actually taken by the spec.

Yep, I referred to those problems in my review as well. For the HMAC (or the usage of the nonce) we may be able to extend the algorithm relatively easily to incorporate those; actually, I believe we have already discussed the alternative at some point. In fact, this privacy analysis might be the very reason why we _should_ make those changes.

As for the re-ordering that is less clear. But... what about the possibility to provide, as a parameter to the algorithm, an alternative sorting function? In most (all?) environments I have worked with the sorting function can be parametrized with such a functional parameter, so this would not create a huge load on the algorithm overall, and might also be a way to de-correlate data to the outside world...

@dlongley what do you think?

spec/index.html
dlongley: It may be better to just make it clear that other specs may parameterize the hash function using whatever they like -- noting that it will not longer be URDNA2015 (or whatever we end up calling it), but instead some variant.

iherman: Naming aside, I would expect all parameters would have a default, probably using SHA256 for hash, no nonce, and sorting just the good old unicode sorting as now. Ie, if we do these additions with defaults we would not harm deployed code (I presume that is your worry, @dlongley). Defining another spec that does exactly the same but parametrizing the hash function sounds a bit of an overkill to me...

dlongley:
To elaborate more on the general issue here:

In order to perform canonicalization, i.e., "make the dataset look the same", we use the data within the dataset itself. We have no other input data. This necessarily means that the output for those things that we canonicalize (i.e., blank node labels and total ordering) will depend on the input.

Notably, this is desirable in all of the VC use cases *as a step* in some transformation process. We want people to be able to write software that when run on machine A or B or that takes in a dataset in format X or Y -- will be able to modify it to get to the same starting point. However, it may be *undesirable* for it to be the *final step* prior to applying cryptography. I think this is the main tripping point and we should highlight this apparent inversion to intuition.

The additional transformation steps often involve reestablishing independence from the input data in how the dataset is expressed -- because this coupling can have undesirable privacy characteristics. This means: blank node labels and the total ordering should be made to *not* depend on the input data (or at least indistinguishably appear not to). Note that this is not just "undoing the work of canonicalization" -- as a baseline has to be established before the work can proceed or  else different software may produce different results.

So, what is needed from our spec to accomplish this, I think, are mappings that connect what is done in the canonicalization algorithm to the original inputs, so algorithm authors can make whatever adjustments they need to. We have also parameterized the internally used hash function to allow that to be replaced -- but, as I noted above, changing this to something like an HMAC will only cause bnode labels to become partially independent from the data in the dataset. I think only applying an HMAC (or similar) function after canonicalization has finished can guarantee full independence.

Similarly, applying a sort order that is based on the input data via a custom comparator (as an example), will also "be operating on the data". Of course, such a comparator could do something like HMAC each N-Quad as they are sorted to produce a pseudorandom but stable ordering -- but this could also be applied to the normalized dataset instead of using the default canonical N-Quads serialization and ordering, without us needing to provide the capability from "within" our canonical serialization description.

Given what I see is a need for reestablishing independence from the data as the driving factor here, I don't think adding even more knobs to be run within the canonicalization algorithm -- which makes its decisions based on the data -- is really helpful. It's better to ensure we can output mappings and don't do too much in a single step ... but I feel this is what we've ended up on already (once issue #89 is addressed).

iherman:
> Given what I see is a need for reestablishing independence from the data as the driving factor here, I don't think adding even more knobs to be run within the canonicalization algorithm -- which makes its decisions based on the data -- is really helpful. It's better to ensure we can output mappings and don't do too much in a single step by separating bnode label computation from canonical serialization so they can be called independently. But ... I feel this is what we've ended up on already (once issue #89 is addressed).

@dlongley I do not have any issues with what you say (not really having an experience in these problems anyway😀). I have come to this from a different angle. The privacy section, as proposed in this PR, seems to suggest that the problems can be solved _within the framework of the current specification_. If we decide not to add more knobs (which is fine with me) then the relevant texts in the proposal (different hashes, usage of nonce or not, sorting) should be formulated to make it clear that a privacy preserving application _must use a slightly different algorithm rather than the one defined by the WG_. The consistency of the specification requires that. On the other, with my W3C staff's hat on, my worry is the the privacy review will pick on this and will require us to make such modifications (ie, adding "knobs") to make the algorithm defined by this WG properly prepared for privacy issues.

At this point I would leave it to you and @yamdan to figure out a proper formulation to possibly clarify this; again, my knowledge and experience on these issues are not fool-proof enough...

spec/index.html
```diff
@@ -2758,13 +2758,13 @@ <h4>Possible Leakage via Canonical Labeling</h4>
       <p>However, in this example, anyone can guess the unrevealed statement by exploiting the canonical identifiers and order.</p>
       
       <p>Since the dataset was sorted in the canonical order, we can get to know that the hidden statement must start with
-        `_:c14n1 &lt;http://schema.org/[f-g]`, which helps us guess that the hidden predicate is
-        `&lt;http://schema.org/gender&gt;` with high probability. Alternatively, we can assume that the guesser already has
+        `_:c14n1 &lt;http://schema.org/[f-g]`, which helps us guess that the hidden predicate is, with high probability,
```
TallTed, line 2761:
```suggestion
        `_:c14n1 &lt;http://schema.org/[f-g]`, which helps us guess with high probability that the hidden predicate is
```

spec/index.html
```diff
@@ -2795,16 +2795,16 @@ <h4>Possible Leakage via Canonical Labeling</h4>
         -->
       </pre>
 
-      <p>Note that we can use the same approach to guess non-boolean values if the range of possible values is still a
-        reasonable (small) size for a guesser to try all the possibilities.</p>
+      <p>Note that we can use the same approach to guess non-boolean values if the range of possible values is still of a
+        reasonably small size for to try all the possibilities.</p>
```
TallTed, line 2799:
```suggestion
      <p>Note that we can use the same approach to guess non-boolean values if the range of possible values is still of a
        reasonably small size, allowing us to try all possibilities.</p>
```

peacekeeper:
We discussed this issue during the [16th May 2023 WG call](https://www.w3.org/events/meetings/493561b0-3a22-467f-8531-3c93ee3512f5/20230516T100000), and there seemed to be consensus that:

1. the WG felt that the content created in this PR by @yamdan was very valuable. (Thanks!)
2. the Privacy Considerations section in rdf-canon should be relatively minimal, for example mention some general privacy impacts that could arise from using this work.
3. but much of the detailed analysis and examples in the PR should be moved "higher up in the spec stack", i.e. to [Data Integrity](https://w3c.github.io/vc-data-integrity/), since many Privacy Considerations have to do less with rdf-canon per se, and more with how rdf-canon is used in specific ways.

So the TODOs are:
- (to address 3. above): Instead of merging this PR into rdf-canon, create an equivalent PR for Data Integrity.
- (to address 2. above): Write a new PR for rdf-canon that mentions some general considerations without going into details of particular applications. (Volunteers?)

yamdan:
> - (to address 2. above): Write a new PR for rdf-canon that mentions some general considerations without going into details of particular applications. (Volunteers?)

I have tried to write a new PR that mentions more general privacy considerations. #101 

> - (to address 3. above): Instead of merging this PR into rdf-canon, create an equivalent PR for Data Integrity.

will do it later...

yamdan:
Discussed [2023-05-24](https://www.w3.org/2023/05/24-rch-minutes.html). We had consensus to close this PR that has been superseded by PR #101, as mentioned in https://github.com/w3c/rdf-canon/pull/99#issuecomment-1550360393.

PR #100, opened 2023-05-15
Adds the "input blank node identifier map"

### LINKED ISSUE #89: Support ordered input dataset or "list of quads" and optional mapping from input indices to output indices

gkellogg:
and a way to initialize it from an input N-Quads document, otherwise uses arbitrary identifiers for each blank node in the input dataset.

Clears up some ambiguity about original blank node identifiers.

Fixes #89.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/100.html" title="Last updated on May 23, 2023, 5:44 PM UTC (a8e7f01)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/100/7bc4143...a8e7f01.html" title="Last updated on May 23, 2023, 5:44 PM UTC (a8e7f01)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -322,7 +322,7 @@
         otherwise, identifiers are assigned arbitrarily for
         each blank node in the input dataset not previously identified.
         <div class="note">Implementations or environments might deal with blank
-        node identifiers more directly, for example, some implementations might
+        node identifiers more directly; for example, some implementations might
         retain blank node identifiers in the parsed or abstract dataset. Implementations
         are expected to reuse these to enable usable mappings between input blank node
         identifiers and output blank node identifiers outside of the algorithm.</div>
--- a/spec/index.html
+++ b/spec/index.html
@@ -317,7 +317,7 @@
         the algorithm.</dd>
       <dt><dfn>input blank node identifier map</dfn></dt>
       <dd>Records any blank node identifiers already assigned to the
-        <a>input datatset</a>.
+        <a>input dataset</a>.
         If the <a>input dataset</a> is provided as an N-Quads document,
         the <a>map</a> relates blank nodes in the abstract <a>input dataset</a>
         to the blank node identiers used within the N-Quads document,
@@ -546,7 +546,7 @@
         create an <a>blank node identifier</a>. It is initialized to
         <code>0</code>.</dd>
       <dt><dfn>issued identifiers map</dfn></dt>
-      <dd>An <a>ordered map</a> that relates <a>blank nodes identifiers</a> to issued identifiers,
+      <dd>An <a>ordered map</a> that relates <a>blank node identifiers</a> to issued identifiers,
         to prevent issuance of more than one new identifier per existing identifier,
         and to allow <a>blank nodes</a> to
         be assigned identifiers some time after issuance.</dd>
--- a/spec/index.html
+++ b/spec/index.html
@@ -958,7 +958,7 @@
           If the <a>input dataset</a> is an N-Quads document,
           parse that document into an dataset,
           retaining any blank node identifiers used within that document
-          in the <a>input blank node identifier map</a>,
+          in the <a>input blank node identifier map</a>;
           otherwise arbitrary identifiers are assigned for each
           blank node.
           <details>
--- a/spec/index.html
+++ b/spec/index.html
@@ -334,7 +334,7 @@
             the <a>input dataset</a>,</li>
           <li>the <a>input blank node identifier map</a> —
             mapping <a>blank nodes</a> in the input dataset to <a>blank node identifiers</a>, and</li>
-          <li>and the <a>canonical issuer</a> —
+          <li>the <a>canonical issuer</a> —
             mapping identifiers in the input dataset to canonical identifiers</li>
         </ul>
         A concrete serialization of a <a>normalized dataset</a> MUST label
--- a/spec/index.html
+++ b/spec/index.html
@@ -328,16 +328,20 @@
         identifiers and output blank node identifiers outside of the algorithm.</div>
       </dd>
       <dt><dfn>normalized dataset</dfn></dt>
-      <dd>A <a>normalized dataset</a> is the combination of an <a>RDF dataset</a>
-        and a <a>map</a> where [=map/keys=]
-        are <a>blank nodes</a> from the dataset
-        and [=map/values=] are the associated canonical <a>blank node identifiers</a>.
+      <dd>A <a>normalized dataset</a> is the combination of the following:
+        <ul>
+          <li>an <a>RDF dataset</a> —
+            the <a>input dataset</a>,</li>
+          <li>the <a>input blank node identifier map</a> —
+            mapping <a>blank nodes</a> in the input dataset to <a>blank node identifiers</a>, and</li>
+          <li>and the <a>canonical issuer</a> —
+            mapping identifiers in the input dataset to canonical identifiers</li>
+        </ul>
         A concrete serialization of a <a>normalized dataset</a> MUST label
-        all <a>blank nodes</a> using these stable <a>blank node identifiers</a>.
-        <p class="ednote">FIXME: use <a>canonical issuer</a> instead of <a>map</a>?</p>
+        all <a>blank nodes</a> using the canonical <a>blank node identifiers</a>.
       </dd>
       <dt><dfn>identifier issuer</dfn></dt>
-      <dd>An identifier issuer is used to issue new <a>blank node identifier</a>. It
+      <dd>An identifier issuer is used to issue new <a>blank node identifiers</a>. It
         maintains a
         <a href="#bn-issuer-state">blank node identifier issuer state</a>.</dd>
       <dt><dfn>hash</dfn></dt>
@@ -594,7 +598,12 @@
       <ol>
         <li id="ca-hl.1"><strong>Initialization</strong>.
           Initialize the state needed for the rest of the algorithm
-          using <a href="#canon-state" class="sectionRef"></a>.</li>
+          using <a href="#canon-state" class="sectionRef"></a>.
+          Also initialize the <a>normalized dataset</a> using the <a>input dataset</a>
+          (which remains immutable)
+          the <a>input blank node identifier map</a>
+          (retaining blank node identifiers from the input, or otherwise assigning them arbitrarily);
+          the <a>canonical issuer</a> is added upon completion of the algorithm.</li>
         <li id="ca-hl.2"><strong>Compute first degree hashes</strong>.
           Compute the first degree hash for each blank node in the dataset using <a href="#hash-1d-quads" class="sectionRef"></a>.</li>
         <li id="ca-hl.3"><strong>Canonically label unique nodes</strong>.
@@ -611,7 +620,9 @@
           If more than one node produces the same N-degree hash,
           the order in which these nodes receive a canonical identifier does not matter.</li>
         <li id="ca-hl.6"><strong>Finish</strong>.
-          Return the <a>serialized canonical form</a> of the <a>normalized dataset</a>.</li>
+          Return the <a>serialized canonical form</a> of the <a>normalized dataset</a>.
+          Alternatively, return the <a>normalized dataset</a> containing
+          the <a>input blank node identifier map</a> and <a>canonical issuer</a>.</li>
       </ol>
     </section>
 
@@ -723,8 +734,8 @@
           as there are no remaining blank nodes without canonical identifiers.</p>
 
         <p><a href="#ca.6">Step 6</a> generates
-          the normalized dataset by replacing blank node identifiers in the original
-          input with their canonical identifiers:</p>
+          the normalized dataset by mapping blank node identifiers in the input dataset
+          with canonical identifiers:</p>
 
         <pre id="ex-ca-unique-normalized-dataset" data-transform="updateExample">
           <!--
@@ -934,9 +945,12 @@
           </tbody>
         </table>
 
-        <p><a href="#ca.6">Step 6</a> generates
-          the normalized dataset by replacing blank node identifiers in the original
-          input with their canonical identifiers:</p>
+        <p><a href="#ca.6">Step 6</a> updates
+          the <a>normalized dataset</a>
+          with the <a>canonical issuer</a>,
+          containing an <a>issued identifiers map</a>
+          mapping blank node identifers from the input dataset
+          to their canonical identifiers:</p>
 
         <pre id="ex-ca-normalized-shared-dataset" data-transform="updateExample">
           <!--
@@ -956,7 +970,7 @@
       <ol id="ca">
         <li id="ca.1">Create the <a>canonicalization state</a>.
           If the <a>input dataset</a> is an N-Quads document,
-          parse that document into an dataset,
+          parse that document into an dataset in the <a>normalized dataset</a>,
           retaining any blank node identifiers used within that document
           in the <a>input blank node identifier map</a>;
           otherwise arbitrary identifiers are assigned for each
@@ -969,10 +983,10 @@
               as well as instantiating a new <a>canonical issuer</a>.</p>
             <p>After this algorithm completes,
               the <a>input blank node identifier map</a> state
-              and / or <a>canonical issuer</a> may be used to
-              correlate blank node identifiers used in the
-              <a>input dataset</a> to those used
-              in the <a>normalized dataset</a>.</p>
+              and <a>canonical issuer</a> may be used to
+              correlate blank nodes used in the
+              <a>input dataset</a> with both their original identifiers,
+              and associated canonical identifiers.</p>
           </details>
         </li>
         <li id="ca.2">For every <a>quad</a> <var>Q</var> in <a>input dataset</a>:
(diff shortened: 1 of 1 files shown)


gkellogg:
We've been pretty fuzzy about identifiers in the input dataset, this is a WIP, but should make it more concrete. If not previously assigned (i.e., by parsing an N-Quads document), it assigned arbitrary identifiers to the input. This does not create any interoperability issues, so there is no need to use any particular algorithm for assigning input blank node identifiers. The `input blank node identifier map` can be extracted after running the algorithm and used along with the `normalized dataset` to correlate input identifiers with the resulting canonical identifiers.

spec/index.html
```diff
@@ -240,7 +240,27 @@ <h2>Uses of Dataset Canonicalization</h2>
...
+    <p>This specification defines an algorithm for creating stable
+      <a>blank node identifiers</a> repeatably for different serializations
+      possibly using individualized <a>blank node identifiers</a>
+      of the same RDF graph (dataset) by grounding each <a>blank node</a>
+      through the nodes to which it is connected,
+      essentially creating <em>Skolem <a>blank node identifiers</a></em>.
```
iherman, line 248:
```suggestion
      through the nodes to which it is connected.
```

I am not sure it is worth referring to Skolem identifiers. It is correct that the term "Skolem" exists in mathematical logic in general but, in the RDF world, we only have the term [Skolem IRIs](https://w3c.github.io/rdf-concepts/spec/#section-skolemization) which _MAY_ use blank node identifiers to replace them with stable IRIs. This is irrelevant for our algorithm, and this remark just confuses the reader imho.

spec/index.html
```diff
@@ -240,7 +240,27 @@ <h2>Uses of Dataset Canonicalization</h2>
...
+      As blank node identifiers can be stable even with other changes to a graph (dataset),
+      in some cases it is possible to compute the difference between two graphs (datasets),
+      for example if changes are made only to ground triples,
+      or if new blank nodes are introduced which do not create an automorphic confusion
+      with other existing blank nodes.
+      If any information which would change the generated blank node identifier,
+      a resulting diff might indicate a greater set of changes than actually exists.
+      Additionally, if the starting dataset is an N-Quads document,
+      which may be based on a previously canonicalized dataset,
+      it may be possible to correlate the original blank node identifiers
```
iherman, line 261:
```suggestion
      Additionally, if the starting dataset is an N-Quads document,
      it may be possible to correlate the original blank node identifiers
```

The fact whether the N-Quads comes from a previously canonicalized dataset or not is irrelevant imho.

spec/index.html
```diff
@@ -295,13 +315,26 @@ <h3>Terms defined by this specification</h3>
...
+      <dt><dfn>input blank node identifier map</dfn></dt>
+      <dd>Records any blank node identifiers already assigned to the
+        <a>input dataset</a>.
+        If the <a>input dataset</a> is provided as an N-Quads document,
+        the <a>map</a> relates blank nodes in the abstract <a>input dataset</a>
+        to the blank node identiers used within the N-Quads document,
+        otherwise, identifiers are assigned arbitrarily for
+        each blank node in the input dataset not previously identified.
+        <div class="note">Some implementations may retain blank node identifiers
+          in the parsed dataset.</div>
```
iherman, line 327:
```suggestion
        <div class="note">Some implementations may retain blank node identifiers
          of the parsed dataset.</div>
```

spec/index.html
```diff
@@ -922,12 +954,23 @@ <h3>Algorithm</h3>
...
+          otherise arbitrary identifiers are assigned for each
+          blank node.
           <details>
             <summary>Explanation</summary>
             <p>This has the effect of initializing the
               <a>blank node to quads map</a>,
               and the <a>hash to blank nodes map</a>,
               as well as instantiating a new <a>canonical issuer</a>.</p>
+            <p>After this algorithm completes,
+              the <a>input blank node identifier map</a> state,
```
iherman, line 970:
```suggestion
              the <a>input blank node identifier map</a> state
```

iherman (commented):
(Github does not allow me to put this comment to an old, unedited area)

Shouldn't §4.4.3/7 make it explicit that the algorithm also return (or return optionally?) The canonical issuer, that will contain the mapping of the input bnode identifiers to the canonical ones?

spec/index.html
```diff
@@ -922,12 +954,23 @@ <h3>Algorithm</h3>
 
       <ol id="ca">
         <li id="ca.1">Create the <a>canonicalization state</a>.
+          If the <a>input dataset</a> is an N-Quads document,
+          parse that document into an dataset,
+          retaining any blank node identifiers used within that document
+          in the <a>input blank node identifier map</a>,
+          otherise arbitrary identifiers are assigned for each
```
yamdan, line 961:
```suggestion
          otherwise arbitrary identifiers are assigned for each
```

yamdan:
Thanks @gkellogg , I think the **input blank node identifier map** introduced here is helpful in practice.
Some N-Quads parsers intentionally randomize blank node identifiers after parsing. For example, in cases where parsers are used when loading N-Quads into a datastore (e.g., [quadstore](https://github.com/belayeng/quadstore/) and [oxigraph](https://github.com/oxigraph/oxigraph)), such randomizations can be taken to avoid blank node identifier collisions. While canonicalization on loading may be rare, my implementation currently uses such a parser, so I have to implement this map later...

spec/index.html
iherman: Is "might concretely deal with" correct in English? It sounds a bit strange to my ears. Can't we just say "might deal with" ?

dlongley: Yeah, it's fine to simplify that -- I removed `concretely`.

dlongley:
@iherman,

> Shouldn't §4.4.3/7 make it explicit that the algorithm also return (or return optionally?) The canonical issuer, that will contain the mapping of the input bnode identifiers to the canonical ones?

Yes, I think it should be explicit that it may optionally return it. However we can do that without modifying the step numbering as @gkellogg mentioned I'd be happy with.

gkellogg:
Another thing to consider:

Update the definition of a __normalized dataset__ to be the input dataset, the __input blank node identifier map__, and the __canonical issuer__. This is useful, as the __input blank node identifier map__ maps blank nodes in the input dataset to associated blank node identifiers, and the __canonical issuer__ maps those identifiers to canonical identifiers. This allows an implementation to either take the N-Quads result, or the __normalized dataset__, without needing to keep this as two separate things.

spec/index.html
```diff
@@ -323,8 +321,11 @@ <h3>Terms defined by this specification</h3>
         to the blank node identiers used within the N-Quads document,
         otherwise, identifiers are assigned arbitrarily for
         each blank node in the input dataset not previously identified.
-        <div class="note">Some implementations may retain blank node identifiers
-          in the parsed dataset.</div>
+        <div class="note">Implementations or environments might deal with blank
+        node identifiers more directly, for example, some implementations might
```
TallTed, line 325:
```suggestion
        node identifiers more directly; for example, some implementations might
```

spec/index.html
```diff
@@ -958,7 +959,7 @@ <h3>Algorithm</h3>
           parse that document into an dataset,
           retaining any blank node identifiers used within that document
           in the <a>input blank node identifier map</a>,
-          otherise arbitrary identifiers are assigned for each
+          otherwise arbitrary identifiers are assigned for each
```
TallTed:
```suggestion
          in the <a>input blank node identifier map</a>;
          otherwise, arbitrary identifiers are assigned for each
```

spec/index.html
```diff
@@ -295,13 +313,29 @@ <h3>Terms defined by this specification</h3>
       <dt><dfn data-lt="input dataset|input datasets">input dataset</dfn></dt>
       <dd>The abstract <a>RDF dataset</a> that is provided as input to
         the algorithm.</dd>
+      <dt><dfn>input blank node identifier map</dfn></dt>
+      <dd>Records any blank node identifiers already assigned to the
+        <a>input dataset</a>.
+        If the <a>input dataset</a> is provided as an N-Quads document,
+        the <a>map</a> relates blank nodes in the abstract <a>input dataset</a>
+        to the blank node identiers used within the N-Quads document,
```
dlongley, line 321:
```suggestion
        to the blank node identifiers used within the N-Quads document,
```

spec/index.html
```diff
@@ -922,12 +955,23 @@ <h3>Algorithm</h3>
...
             <summary>Explanation</summary>
             <p>This has the effect of initializing the
               <a>blank node to quads map</a>,
               and the <a>hash to blank nodes map</a>,
               as well as instantiating a new <a>canonical issuer</a>.</p>
+            <p>After this algorithm completes,
+              the <a>input blank node identifier map</a> state
+              may be used to correlate blank node identifiers
+              used in the <a>input dataset</a> to those used
+              in the <a>normalized dataset</a>.</p>
```
dlongley, line 974: We need to expose the input blank node ID => canonical blank node ID ordered map from the canonical issuer as well -- as it's the easiest thing for some implementations to use:

```suggestion
            <p>After this algorithm completes,
              the <a>input blank node identifier map</a> state
              and / or <a>canonical issuer</a> may be used to
              correlate blank node identifiers used in the
              <a>input dataset</a> to those used
              in the <a>normalized dataset</a>.</p>
```

I see there's a comment from @gkellogg that perhaps we want to expose the canonical issuer (or its input bnode ID => output bnode ID) map in the normalized dataset instead of a input bnode => output bnode ID map ... and that may simplify this here without losing anything.

If the "input blank node identifier map" always maps the abstract bnodes in the input dataset (and, therefore the normalized dataset as well) to the input blank node IDs, then the canonical issuer's ordered map can:

* be used to map from input blank node IDs to output blank node IDs (this is a single step), and
* it can be used in conjunction with the "input blank node identifier map" to go, in two steps, from bnodes to input bnode IDs to output node IDs.

spec/index.html
```diff
@@ -969,8 +969,9 @@ <h3>Algorithm</h3>
               as well as instantiating a new <a>canonical issuer</a>.</p>
             <p>After this algorithm completes,
               the <a>input blank node identifier map</a> state
-              may be used to correlate blank node identifiers
-              used in the <a>input dataset</a> to those used
+              and / or <a>canonical issuer</a> may be used to
```
TallTed, line 972:
```suggestion
              and/or <a>canonical issuer</a> may be used to
```

gkellogg:
Last commit adds the __input blank node identifier map__ and __canonical issuer__ to the __normalized dataset__. This makes the RDF dataset portion of this immutable, initialized from the __input dataset__.

Updates some steps to clarify that updates are for map entries, and not adding quads, which simplifies step 6.

spec/index.html
gkellogg: That text was changed, as it now includes both these things.

spec/index.html
```diff
@@ -1255,7 +1311,25 @@ <h3>Algorithm</h3>
...
+              from the canonical identifiers associated with each blank node.</p>
+            <p>The <a>normalized dataset</a> is composed of the original
+              <a>input dataset</a>, the <a>input blank node identifier map</a>,
+              containing identifiers for each blank node in the <a>input dataset</a>,
+              and the <a>canonical issuer</a>,
+              containing an <a>issued identifiers map</a>
+              mapping the identifiers in the <a>input blank node identifier map</a>
+              to their canonical identifiers.
+            </p>
+          </details>
```
yamdan, line 1331: I think it would be better to note that the *normalized dataset* should not be interpreted as a single canonical representation because the algorithm can output different canonical issuers depending on the implementation or runtime environment. (https://github.com/w3c/rdf-canon/issues/89#issuecomment-1538275081)

For example, an input dataset
```
_:e0 <http://example.org/vocab#next> _:e1 .
_:e1 <http://example.org/vocab#next> _:e0 .
```
can be transformed into the normalized dataset with either one of the following canonical issuers, depending on the implementation:
1. `{ "e0": "c14n0", "e1": "c14n1" }`
2. `{ "e0": "c14n1", "e1": "c14n0" }`

Both canonical issuers result in the same single serialized form:
```
_:c14n0 <http://example.org/vocab#next> _:c14n1 .
_:c14n1 <http://example.org/vocab#next> _:c14n0 .
```
So, we can only say that serialized form is a single canonical representation, but the normalized dataset is possibly not.

gkellogg: It doesn't say that the normalized dataset is a single canonical representation; as you point out, the association of blank nodes to input identifiers could be different for two otherwise isomorphic datasets, and therefor the map from input identifier to canonical identifier would differ. Note that this is in a non-normative explanation detail. Is there some specific text you'd like to add or change?

spec/index.html
```diff
@@ -1224,21 +1286,15 @@ <h3>Algorithm</h3>
...
+        <li id="ca.6">Add the <a>canonical issuer</a> to the
+          <a>normalized dataset</a>.
           <details>
             <summary>Explanation</summary>
-            <p>This step populates the <a>normalized dataset</a> with quads
-              substituting the original blank node identifiers,
-              with the newly established canonical blank node identifiers.</p>
+            <p>This step adds the <a>canonical issuer</a> to the
+              <a>normalized dataset</a>, the [= map/key | keys =] in the
+              <a>canonical issuer</a> with be [= map/entry | map entries =] of the
```
dlongley, line 1295:
```suggestion
              <a>canonical issuer</a> with the [= map/entry | map entries =] of the
```

dlongley (approved):
Great, thanks! Approving now -- and I have no issue if there's some informative note added around the isomorphic datasets discussion between @yamdan and @gkellogg.

spec/index.html
```diff
@@ -1255,7 +1311,25 @@ <h3>Algorithm</h3>
           </details>
         </li>
         <li id="ca.7">Return the <a>serialized canonical form</a>
-          of the <a>normalized dataset</a>.</li>
+          of the <a>normalized dataset</a>.
+          Alternatively, return the <a>normalized dataset</a> itself,
+          which includes the <a>input blank node identifier map</a>,
+          and <a>canonical issuer</a>.
```
yamdan, line 1317:
```suggestion
          Optionally, the algorithm may also return the <a>normalized dataset</a> as an auxiliary output,
          which includes the <a>input blank node identifier map</a>,
          and <a>canonical issuer</a>.
```
How about saying that the *output* of the c14n (single and deterministic) is the serialized form, whereas the normalized dataset (possibly non-deterministic) can be obtained as an *auxiliary output*?

gkellogg: So, would change "Optionally" to "As an auxiliary output" satisfy this? I don't really follow how this is non-deterministic, as it would seem that with a given input, the same normalized dataset would be produced.

dlongley: I'd rather not be so prescriptive in saying that you have to return both -- I'd be happy with an either / or. I don't think we should say what implementations can do as a proxy for indicating that two different implementations might technically map one blank node to ID A and another implementation might map it to ID B. This is @yamdan's point, I believe -- but this only happens when there are isomorphisms that make this difference irrelevant. A serialized version of the dataset would look the same.

Perhaps that's what we say in a note: "Technically speaking, one implementation might map particular blank nodes to different identifiers than another implementation, however, this only occurs when there are isomorphisms in the dataset such that a serialized expression of the dataset would appear the same from either implementation."

And then we can say that algorithms may return both the canonically serialized dataset and the normalized dataset or either of these as requested by the invoker of the algorithm.

gkellogg: It might be some time before I can do this update, but it seems simple enough. I'm traveling for the next week, and internet access is spotty. Feel free to update and commit, as this is really just informative, now.

dlongley, line 1317: Here's a suggestion to try and address the concerns here:

```suggestion
          Alternatively, or additionally, an implementation may return,
          upon request, the <a>normalized dataset</a> itself, which
          includes the <a>input blank node identifier map</a>,
          and <a>canonical issuer</a>.
          <p class="note">Technically speaking, one implementation
          might return a <a>normalized dataset</a> that maps
          particular blank nodes to different identifiers than another
          implementation, however, this only occurs when there are
          isomorphisms in the dataset such that a canonically serialized
          expression of the dataset would appear the same from either
          implementation.</p>
```

dlongley: I've added a suggestion below.

yamdan: @gkellogg,
> I don't really follow how this is non-deterministic, as it would seem that with a given input, the same normalized dataset would be produced.

**Non-deterministic** choice may occur in step 5.3 of the [4.4.3 Algorithm](https://w3c.github.io/rdf-canon/spec/#canon-algo-algo), where it's possible to have ties of *result* in the *hash path list* so that it's non-deterministic which *result* is firstly chosen from the list.
Even the same implementation can output different canonical issuers depending on the runtime environment or the input blank node identifiers.

The only thing I would like to eliminate here is the possibility of creating a misuse of the *normalized dataset*, believing that it is a deterministic and single canonical result and connecting it to the hash or signature input.
I think we can prevent this by clearly stating that the serialized form is the output of the canonicalization and the normalized dataset is an auxiliary output.

As @dlongley mentioned, I think this only happens when there are automorphisms in the input dataset.

spec/index.html
```diff
@@ -561,7 +598,12 @@ <h3>Overview</h3>
       <ol>
         <li id="ca-hl.1"><strong>Initialization</strong>.
           Initialize the state needed for the rest of the algorithm
-          using <a href="#canon-state" class="sectionRef"></a>.</li>
+          using <a href="#canon-state" class="sectionRef"></a>.
+          Also initialize the <a>normalized dataset</a> using the <a>input dataset</a>
+          (which remains immutable)
+          the <a>input blank node identifier map</a>
+          (retaining blank node identifiers from the input, or otherwise assigning them arbitrarily);
```
dlongley, line 605:
```suggestion
          (retaining blank node identifiers from the input if possible, otherwise assigning them arbitrarily);
```

spec/index.html
yamdan: I will continue this "non-deterministic" topic in a new separate PR.

yamdan:
We decided to merge this PR on 2023-05-24 WG Call. I will make new PR related to "non-deterministic" canonical issuer topic.

PR #101, opened 2023-05-19
Add more general privacy considerations 

yamdan:
Based on https://github.com/w3c/rdf-canon/pull/99#issuecomment-1550360393 and the discussion on the [16th May 2023 WG call](https://www.w3.org/2023/05/16-rch-minutes.html), I have rewritten the privacy consideration to be more general and lightweight. The reference to Data Integrity spec is not valid at this time, so it might be better to rewrite it. I would appreciate your comments and suggestions...

Supersedes #99.
Fixes #84.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/101.html" title="Last updated on May 20, 2023, 2:23 AM UTC (66a6878)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/101/7bc4143...66a6878.html" title="Last updated on May 20, 2023, 2:23 AM UTC (66a6878)">Diff</a>

PR #102, opened 2023-05-19
Remove "IANA Considerations"

yamdan:
Based on @iherman's comment (https://github.com/w3c/rdf-canon/pull/101#pullrequestreview-1434216805);
I think "IANA Considerations" should be moved from Security Considerations as another independent section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/102.html" title="Last updated on May 20, 2023, 2:24 AM UTC (cf9bbd4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/102/7bc4143...cf9bbd4.html" title="Last updated on May 20, 2023, 2:24 AM UTC (cf9bbd4)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2707,13 +2707,6 @@ attacker sends a poisoned dataset might be different from system to system.
     </p>
   </section>
 
-
-  <section id="iana">
-    <h2>IANA Considerations</h2>
-
-    <p class="ednote">TODO</p>
-  </section>
-
   <section>
     <h3>Formal Verification Incomplete</h3>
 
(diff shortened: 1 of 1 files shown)


iherman:
> Difintitely should be on its own, but I’m not sure why we need a section at all, as we don’t define a. Edina type, nor even extend one. N-Quads defines the media type.

I must admit I did not really looked at the (empty) section... And indeed, we do not define a media type, so the simplest thing would be to remove that section altogether.

yamdan:
Thank you for your comments. I restate this PR to *remove* IANA Considerations.

gkellogg (approved):

dlongley (approved):

Issue #70, opened 2023-01-20
Security considerations


philarcher:
This just a place where we can discuss any security issues that we think might arise from the spec. Looking at the [self-assessment form](https://www.w3.org/TR/security-privacy-questionnaire/) it doesn't look like we will have much to worry about but we need to consider this carefully.

philarcher:
This was discussed 2023-09-11 (see below https://www.w3.org/2023/09/11-rch-minutes.html#t05). Group feels the issue has been addressed fully. Text to include new reference to the Data Integrity spec for more info.

PR #106, opened 2023-05-24
Add security considerations text.

dlongley:
Addresses #70.

This PR removes the issue text in the Security Considerations section and adds text on dataset poisoning. The formal verification text is removed as a non-requirement for this section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/106.html" title="Last updated on Jun 7, 2023, 2:28 PM UTC (55c2aa4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/106/6d2e77f...55c2aa4.html" title="Last updated on Jun 7, 2023, 2:28 PM UTC (55c2aa4)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2791,10 +2791,10 @@ disclose.
       information connected to blank nodes in order to ensure that each will
       properly receive its own canonical identifier. This process can be
       exploited by attackers to construct datasets which are known to take
-      large amounts of computing time to canonize, but that do not express
+      large amounts of computing time to canonicalize, but that do not express
       useful information or express it using unnecessary complexity.
       Implementers of the algorithm are expected to add mitigations that will,
-      by default, abort canonizing problematic inputs.
+      by default, abort canonicalizing problematic inputs.
     </p>
     <p>Suggested mitigations include, but are not limited to:
       <ul>
--- a/spec/index.html
+++ b/spec/index.html
@@ -2787,31 +2787,29 @@ disclose.
   <section>
     <h3>Dataset Poisoning</h3>
 
-    <p class="issue" data-number="70" title="Attackers can construct poison datasets">
-Add text that warns that attackers can construct datasets which are known to
-take large amounts of compute time to canonize. The algorithm has a mechanism to detect
-and prevent this sort of abuse, but implementers need to ensure that they
-think holistically about their system such as what happens if they don't have
-rate limiting on a service exposed to the Internet and they are the subject of
-a DDoS. Default mechanisms that prevent excessive use of compute when an
-attacker sends a poisoned dataset might be different from system to system.
+    <p>The canonicalization algorithm examines every difference in the
+      information connected to blank nodes in order to ensure that each will
+      properly receive its own canonical identifier. This process can be
+      exploited by attackers to construct datasets which are known to take
+      large amounts of compute time to canonize, but that do not express
+      useful information or express it using unnecessary complexity.
+      Implementers of the algorithm are expected to add mitigations that will,
+      by default, abort canonizing problematic inputs.
     </p>
-  </section>
-
-  <section>
-    <h3>Formal Verification Incomplete</h3>
-
-    <p class="issue" data-number="70" title="Formal verification of algorithm is incomplete">
-Add text that warns implementers that, while the algorithm has a mathematical
-proof associated with it that has had peer review, and while a W3C WG
-has reviewed the algorithms and that there are multiple interoperable
-implementations, that a formal proof using a system such as Coq isn't available
-at this time. We are highly confident of the correctness of the algorithm,
-but we will not be able to say with 100% certainty that it is correct until
-we have a formal, machine-based verification of the proof. Any system that
-utilizes this canonicalization mechanism should have a backup canonicalization
-mechanism, such as JCS, or other mitigations, such as schema-based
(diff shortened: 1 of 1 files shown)


gkellogg (approved):

iherman (approved):
- I wonder whether it would be useful to suggest some values (based on current experiences) for the number of recursions and/or a reasonable timeout value for the suggested methods. Just to help implementers to start somewhere...
- (for a future editorial review) _If_ dataset poisoning is the only security issue we refer to, then we may not want to have a separate subsection. It is all right to keep it as is for now, but just keep in mind when we publish the final version.
-

spec/index.html
```diff
@@ -2787,31 +2787,29 @@ <h2>Security Considerations</h2>
...
-and prevent this sort of abuse, but implementers need to ensure that they
-think holistically about their system such as what happens if they don't have
-rate limiting on a service exposed to the Internet and they are the subject of
-a DDoS. Default mechanisms that prevent excessive use of compute when an
-attacker sends a poisoned dataset might be different from system to system.
+    <p>The canonicalization algorithm examines every difference in the
+      information connected to blank nodes in order to ensure that each will
+      properly receive its own canonical identifier. This process can be
+      exploited by attackers to construct datasets which are known to take
+      large amounts of compute time to canonize, but that do not express
```
yamdan, line 2794:
```suggestion
      large amounts of computing time to canonize, but that do not express
```

yamdan (approved):
Thank you @dlongley!

dlongley:
@iherman,

> I wonder whether it would be useful to suggest some values (based on current experiences) for the number of recursions and/or a reasonable timeout value for the suggested methods. Just to help implementers to start somewhere...

I thought about that -- but felt like it could vary depending on a number of factors to the point that saying something like the default timeout should be "on the order of a few milliseconds" or the max deep recursions should be "probably 1 for common use" felt wishy-washy.

I think what ecosystem a library was primarily written for (whether it's very general or more specific), whether it runs in a single thread or multiple threads (or has features to allow for other work scheduling), whether its target computing devices have X power or Y, and so on all impact the default values. The defaults are about what the predicted most common use cases will be. Ultimately, it's just about making the canonicalization algorithm run in some "reasonable amount of time" or some "reasonable about of compute" for their biggest user base -- which an implementer needs to figure out.

spec/index.html
```diff
@@ -2787,31 +2787,29 @@ <h2>Security Considerations</h2>
...
-we have a formal, machine-based verification of the proof. Any system that
-utilizes this canonicalization mechanism should have a backup canonicalization
-mechanism, such as JCS, or other mitigations, such as schema-based
-validation, ready in the event that an unrecoverable flaw is found in this algorithm.
+    <p>Suggested mitigations include, but are not limited to:
+      <ul>
+        <li>providing a configurable timeout with a default value applicable to
+          an implementation's common use, and / or</li>
+        <li>providing a configurable limit on the number of iterations of steps
+          performed in the algorithm, particularly recursive steps.</li>
```
TallTed, line 2804: This list would be improved by adding a third mitigation, even if it's minimally effective.
```suggestion
          an implementation's common use</li>
        <li>providing a configurable limit on the number of iterations of steps
          performed in the algorithm, particularly recursive steps</li>
```

dlongley: > This list would be improved by adding a third mitigation, even if it's minimally effective.

Hmm, I don't have one in mind -- and I hesitate to add one just because three reads better than two to most humans. It's also the case that the following paragraph spells out a third mitigation, but one that is to be performed (and generally automatically is) at the application layer.

philarcher:
Thanks for doing this Dave.

One thing: does canonize mean the same thing as canonicalize in US English? You've used 'canonize' a couple of times which, in my version of English, means the algorithm creates a Saint Dataset (i.e. to canonize is to make someone a saint). See https://www.merriam-webster.com/dictionary/canonize

TallTed:
I believe we should use `canonicalize` in preference to `canonize`, throughout.

spec/index.html
```diff
@@ -2787,31 +2787,29 @@ <h2>Security Considerations</h2>
...
-and prevent this sort of abuse, but implementers need to ensure that they
-think holistically about their system such as what happens if they don't have
-rate limiting on a service exposed to the Internet and they are the subject of
-a DDoS. Default mechanisms that prevent excessive use of compute when an
-attacker sends a poisoned dataset might be different from system to system.
+    <p>The canonicalization algorithm examines every difference in the
+      information connected to blank nodes in order to ensure that each will
+      properly receive its own canonical identifier. This process can be
+      exploited by attackers to construct datasets which are known to take
+      large amounts of computing time to canonize, but that do not express
```
dlongley, line 2794:
```suggestion
      large amounts of computing time to canonicalize, but that do not express
```

spec/index.html
```diff
@@ -2787,31 +2787,29 @@ <h2>Security Considerations</h2>
...
+      useful information or express it using unnecessary complexity.
+      Implementers of the algorithm are expected to add mitigations that will,
+      by default, abort canonizing problematic inputs.
```
dlongley, line 2797:
```suggestion
      by default, abort canonicalizing problematic inputs.
```

dlongley (commented):
Commenting to use suggestion feature.

PR #137, opened 2023-07-05
Updated rdfjs-c14n-report

iherman:
This update covers the latest test suite, including the poison graph

### DIFF
--- a/reports/earl.jsonld
+++ b/reports/earl.jsonld
@@ -135,7 +135,7 @@
       "name": "earl-report-0.8.0",
       "doap:created": {
         "@type": "http://www.w3.org/2001/XMLSchema#date",
-        "@value": "2023-07-04"
+        "@value": "2023-07-05"
       },
       "revision": "0.8.0"
     },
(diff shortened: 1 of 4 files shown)


dlongley (approved):

gkellogg:
The way to test for this is to run the `rake` task in the reports directory after updating an EARL submission; it will report on any errors. IN this case, the test identifiers don't match up with those in the test manifest (e.g., `https://w3c.github.io/rdf-canon/tests/#test074c` instead of `https://w3c.github.io/rdf-canon/tests/manifest#test074c`.

I'll correct so it will re-run.

gkellogg (approved):

iherman:
> The way to test for this is to run the `rake` task in the reports directory after updating an EARL submission; it will report on any errors. IN this case, the test identifiers don't match up with those in the test manifest (e.g., `https://w3c.github.io/rdf-canon/tests/#test074c` instead of `https://w3c.github.io/rdf-canon/tests/manifest#test074c`.
> 
> I'll correct so it will re-run.

Oops, sorry about that, @gkellogg. I have updated my tester software, it should be o.k. next time.

PR #138, opened 2023-07-12
Add normative requirement to prevent denial-of-service attacks

### LINKED ISSUE #120: What a reasonable way of defending against poison graphs?
This is more a question to those who have a practical experience of the algorithm implementation out there, ie, @davidlehn @dlongley and maybe @gkellogg.

As a protection against poison graphs what I did in my implementation is as follows:

1. I have set a maximum limit to the recursion in [5.4.5.1.](https://www.w3.org/TR/rdf-canon/#hndq.5.4.5.1). I.e., if the "level" of the recursion has reached a limit, I stop the algorithm. At this moment, the limit is set to 50, but I must admit this more a gut feeling rather than anything else.
2. The user has the possibility, via a separate call in the interface, to set a _lower_ maximum limit (but cannot exceed that limit).

Is this a reasonable approach? If yes, is 50 indeed a reasonable limit?

An alternative would be to let the system-wide limit be set at positive infinity (more exactly, the maximum integer representable in JavaScript), and the user can set a limit by herself.

Any advice/experience is welcome...

gkellogg:
... via poison datasets.

Fixes #120.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/138.html" title="Last updated on Jul 14, 2023, 3:19 PM UTC (9098bfe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/138/5c8e5e6...9098bfe.html" title="Last updated on Jul 14, 2023, 3:19 PM UTC (9098bfe)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -989,7 +989,7 @@
       <p class="note">Implementations can consider placing limits on the number of
         calls to <a href="#hash-nd-quads"></a> based on the number
         of blank nodes in the <a>hash to blank nodes map</a>.
-        For most well-formed datasets, no more than a couple
+        For most well-formed datasets, more than a couple
         of iterations on <a href="#hash-nd-quads"></a> per blank node would be unusual.</p>
 
       <ol id="ca">
--- a/spec/index.html
+++ b/spec/index.html
@@ -981,7 +981,7 @@
       <h3>Algorithm</h3>
 
       <p>The following algorithm will run with a minimal number of iterations in each step
-        for well formed <a>input datasets</a>.
+        for typical <a>input datasets</a>.
         In some extreme cases, the algorihm can behave poorly, particularly in <a href="#ca.5">Step 5</a>.
         Implementations MUST prevent against potenial denial-of-service attacks.
         See <a href="#dataset-poisoning"></a> for futher information.</p>
@@ -989,7 +989,7 @@
       <p class="note">Implementations can consider placing limits on the number of
         calls to <a href="#hash-nd-quads"></a> based on the number
         of blank nodes in the <a>hash to blank nodes map</a>.
-        For most well-formed datasets, more than a couple
+        For most typical datasets, more than a couple
         of iterations on <a href="#hash-nd-quads"></a> per blank node would be unusual.</p>
 
       <ol id="ca">
--- a/spec/index.html
+++ b/spec/index.html
@@ -980,6 +980,18 @@
     <section id="canon-algo-algo">
       <h3>Algorithm</h3>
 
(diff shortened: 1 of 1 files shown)


gkellogg:
I think this satisfies the requirement without being overly prescriptive.

spec/index.html
```diff
@@ -980,6 +980,18 @@ <h3>Examples</h3>
     <section id="canon-algo-algo">
       <h3>Algorithm</h3>
 
+      <p>The following algorithm will run with a minimal number of iterations in each step
+        for well formed <a>input datasets</a>.
```
iherman, line 984: What does "well formed" means? Poison sets are also well formed, i.e., they are bona fide RDF graph datasets...

"average"? "general"? "practical?"

spec/index.html
```diff
@@ -980,6 +980,18 @@ <h3>Examples</h3>
...
+        In some extreme cases, the algorihm can behave poorly, particularly in <a href="#ca.5">Step 5</a>.
+        Implementations MUST prevent against potenial denial-of-service attacks.
+        See <a href="#dataset-poisoning"></a> for futher information.</p>
+
+      <p class="note">Implementations can consider placing limits on the number of
+        calls to <a href="#hash-nd-quads"></a> based on the number
+        of blank nodes in the <a>hash to blank nodes map</a>.
+        For most well-formed datasets, no more than a couple
+        of iterations on <a href="#hash-nd-quads"></a> per blank node would be unusual.</p>
```
iherman, line 993:
```suggestion
        For most well-formed datasets, more than a couple
        of iterations on <a href="#hash-nd-quads"></a> per blank node would be unusual.</p>
```

Note also my comment above on the term "well-formed".

(Minor: even if the term is retained, it should be consistent on whether it is "well-formed" or "well formed".)

iherman (changes requested):

spec/index.html
gkellogg: Maybe. “Well behaved” or “typical”.

iherman: I'd vote for "typical"

TallTed (changes requested):
I would change both instances of `well[- ]formed` to `typical`.

dlongley (approved):
+1 with the suggestions offered by others.

spec/index.html
gkellogg:
```suggestion
        for typical <a>input datasets</a>.
```

iherman (approved):

PR #154, opened 2023-07-27
Turn off hljs on logging pre sections.

### LINKED ISSUE #147: "Logging" expandos have empty context
The "Logging" content is in the spec in commented sections, but show up as empty box in the spec.

Is a more in depth explanation needed for how to interpret that logging content?

gkellogg:
In Chrome, this caused the exxpected log messages not to appear.

Fixes #147.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/154.html" title="Last updated on Jul 28, 2023, 6:21 PM UTC (b9d544c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/154/84315eb...b9d544c.html" title="Last updated on Jul 28, 2023, 6:21 PM UTC (b9d544c)">Diff</a>

### DIFF
--- a/spec/common/typographical-conventions.html
+++ b/spec/common/typographical-conventions.html
@@ -48,7 +48,7 @@
       <summary>Logging</summary>
       <p>For example, the following output snippet might
         describe the operation of an implementation using the [[?YAML]] format.</p>
-      <pre class="logging yaml" data-transform="updateExample">
+      <pre class="logging yaml nohighlight" data-transform="updateExample">
         <!--
         ca:
           ca2:
--- a/spec/index.html
+++ b/spec/index.html
@@ -1051,7 +1051,7 @@
           <details>
             <summary>Logging</summary>
             <p>Log the state of the <a>blank node to quads map</a>:</p>
-            <pre class="logging yaml" data-transform="updateExample">
+            <pre class="logging yaml nohighlight" data-transform="updateExample">
               <!--
                 # Blank node to quads map for unique hashes example
                 ca:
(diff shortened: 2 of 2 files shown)


dlongley (approved):
Works for me.

davidlehn:
- Also needs the fix in `common/typographical-conventions.html`.
- May be worth looking into why YAML-LD spec shows it's `<pre class="example yaml" ...>` text ok, but this logging yaml text doesn't work.  It's not highlighted in YAML-LD as YAML, but it at least shows up.
- Related to both specs is YAML is not a [default supported language](https://respec.org/docs/#pre-and-code-elements).  Looks like another highlighter could be added, but their example uses a CJS script and the highlightjs yaml code is ESM.  Maybe not worth the trouble to figure that out right now.
- The text flush to the border could be improved.  How about in `pre.logging` style adding `padding: 0.3em;`?

yamdan (approved):
Approving with David's comment. I usually use Firefox and didn't notice the issue.

gkellogg:
Highlighter is part of ReSpec, probably don’t want to try to include our own. We use examples that aren’t pure YAML, anyway.

good question on why YAML-LD spec doesn’t show the issue, or why it worked on Safari.

gkellogg:
According to the [highlight.js page](https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md), YAML is a default supported language. But, there's something in ReSpec which is not rendering the result. It does pass through `updateExample` okay, but get's dropped when rendering in ReSpec. I'd consider that a bug, but don't really want to spend the time to track it down right now.

davidlehn:
highlight.js itself supports many languages, but as the respec link I noted above says, it only enables a few by default.  I didn't dig into it enough to figure how to pull in YAML support without too much pain.  May be easier to lobby repsec to include it if other projects need YAML support too.
https://github.com/w3c/respec/blob/main/worker/respec-highlight.js

davidlehn (approved):

PR #159, opened 2023-08-18
Add security section about insecure hashing algorithms.

peacekeeper:
An attempt to add a note about hashing algorithms becoming insecure, to address one of the comments in the TAG review in https://github.com/w3ctag/design-reviews/issues/855#issuecomment-1664350113.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -2884,6 +2884,17 @@ disclose.
     </p>
   </section>
 
+  <section id="insecure-hashing-algorithms" class="informative">
+    <h3>Insecure Hashing Algorithms</h3>
+
+    <p>It is possible that the hashing algorithm used by RDFC-1.0 may become
+      insecure at some point in the future. To mitigate this, this algorithm
+      and implementations of it can be parameterized to use a different
+      hashing function, without the need to make any changes to the
+      canonicalization algorithm itself.
+    </p>
+  </section>
+
 </section>
 
 <section id="use-cases" class="informative">


spec/index.html
```diff
@@ -2884,6 +2884,17 @@ <h3>Dataset Poisoning</h3>
     </p>
   </section>
 
+  <section id="insecure-hashing-algorithms" class="informative">
+    <h3>Insecure Hashing Algorithms</h3>
+
+    <p>It is possible that the hashing algorithm used by RDFC-1.0 may become
```
gkellogg, line 2890: Referencing the definition of the hash algorithm would be useful, as it also makes this statement. (Alternatively, use the reference as "hash algorithm" directly, instead of "hashing algorithm").

```suggestion
    <p>It is possible that the <a data-lt="hash algorithm">hashing algorithm</a> used by RDFC-1.0 may become
```

TallTed: +1 for "hash algorithm", better than "hashing algorithm"

gkellogg: > +1 for "hash algorithm", better than "hashing algorithm"

I updated the suggestion accordingly.

gkellogg (changes requested):
Once my suggestion is acted on, it's good to go. Thanks!

spec/index.html
dlongley: "may" => "might" to avoid normative language.

dlongley (approved):

spec/index.html
```diff
@@ -2884,6 +2884,17 @@ <h3>Dataset Poisoning</h3>
...
+      insecure at some point in the future. To mitigate this, this algorithm
+      and implementations of it can be parameterized to use a different
+      hashing function, without the need to make any changes to the
```
yamdan, line 2893:
```suggestion
      hash function, without the need to make any changes to the
```

spec/index.html
```diff
@@ -2884,6 +2884,17 @@ <h3>Dataset Poisoning</h3>
     </p>
   </section>
 
+  <section id="insecure-hashing-algorithms" class="informative">
+    <h3>Insecure Hashing Algorithms</h3>
```
yamdan, line 2888:
```suggestion
  <section id="insecure-hash-algorithms" class="informative">
    <h3>Insecure Hash Algorithms</h3>
```
in line with @TallTed 's suggestion.

yamdan (approved):
Thanks!

gkellogg (approved):

PR #161, opened 2023-08-30
Parameterized hash

### LINKED ISSUE #160: Parameterized hash algorithm
Some provision, beyond the existing note, should be added to the spec that describes SHA-256 as the default algorithm, and explicitly allows other algorithms to be specified through suitable parameterization.

gkellogg:
* Make the SHA-256 hash algorithm default, and specifically allow for a different algorithm to be substituted.
* Add test vocabulary for specifying hashAlgorithm, and add test075 using the algorithm SHA384.

Fixes #160

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/161.html" title="Last updated on Sep 1, 2023, 5:33 PM UTC (2b45776)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/161/51312d0...2b45776.html" title="Last updated on Sep 1, 2023, 5:33 PM UTC (2b45776)">Diff</a>


spec/index.html
```diff
@@ -370,11 +370,11 @@ <h3>Terms defined by this specification</h3>
...
-      <dd>The hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256.
-        <div class="note">Implementations can be written to parameterize the
-        hash algorithm without any other changes. However, using a different
-        hash algorithm is expected to generate different output from
-        <a>RDFC-1.0</a>.</div>
+      <dd>The default hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256.
+        <p>Implementations can be written to parameterize the
+          hash algorithm without any other changes. However, using a different
+          hash algorithm is expected to generate different output from
+          <a>RDFC-1.0</a>.</p>
```
dlongley, line 377:
```suggestion
          hash algorithm without any other changes. Using a different
          hash algorithm is expected to generate different output from
          using the default.</p>
```

dlongley (approved):
Approving with a very minor tweak suggested. Thanks!

tests/manifest.jsonld
```diff
@@ -13,6 +13,7 @@
       "@id": "mf:action",
       "@type": "@id"
     },
+    "algorithm": "rdfc:hashAlgorithm",
```
davidlehn, line 16: Perhaps better to name `hashAlgorithm` to avoid confusion with the main RDFC-1.0 algorithm

spec/index.html
```diff
@@ -370,11 +370,11 @@ <h3>Terms defined by this specification</h3>
...
-      <dd>The hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256.
-        <div class="note">Implementations can be written to parameterize the
-        hash algorithm without any other changes. However, using a different
-        hash algorithm is expected to generate different output from
-        <a>RDFC-1.0</a>.</div>
+      <dd>The default hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256.
+        <p>Implementations can be written to parameterize the
+          hash algorithm without any other changes. Using a different
+          hash algorithm is expected to generate different output from
+          using the default.</p>
```
davidlehn, line 377: I think here and in the Insecure Hash Algorithm section it needs to be said in stronger words that the canonical output depends on the hash algorithm used.

tests/manifest.jsonld
```diff
@@ -854,6 +855,28 @@
       "computationalComplexity": "high",
       "approval": "rdft:Proposed",
       "action": "rdfc10/test074-in.nq"
+    },
+    {
+      "id": "#test075c",
+      "type": "rdfc:RDFC10EvalTest",
+      "name": "poison – evil (1)",
+      "comment": "A poison graph which is computable given defined limits. (Uses SHA-384)",
```
davidlehn, line 863: Is "computable given defined limits" a poison graph?  This PR is mainly about the hash algorithm, is a poison graph needed here?  Perhaps have simple input tests for this purpose.  Two tests with the same input, different only by specified hash algorithm, and that result in different outputs (pick an input that does differ).  Trying not to mix test purposes here.

davidlehn:
I'm wondering if the hash algorithm will commonly need to be encoded in an algorithm name string, similar to what some other crypto specs [use](https://datatracker.ietf.org/doc/html/rfc7518#section-3.1).  For any use case that supports different hash algorithms, it will need to be communicated somehow.  Where should this sort of issue be handled?  Similar to "HMAC using SHA-256", we have the use case here of "RDFC-1.0 using SHA-256".  Which could verbosely be encoded like `RDFC-1.0-SHA256`.

tests/manifest.csv
```diff
@@ -1,63 +1,64 @@
...
+test066,"blank node - double circle of 3 (1-0-2, reversed)",,3,,,TRUE,FALSE
+test067,"blank node - double circle of 3 (1-2-0, reversed)",,3,,,TRUE,FALSE
+test068,"blank node - double circle of 3 (2-1-0, reversed)",,3,,,TRUE,FALSE
+test069,"blank node - double circle of 3 (2-0-1, reversed)",,3,,,TRUE,FALSE
+test070,dataset - isomorphic default and iri named,Isomorphic graphs in default and IRI named graph,0,,,TRUE,TRUE
+test071,dataset - isomorphic default and node named,Isomorphic graphs in default and blank node named graph,0,,,TRUE,TRUE
+test072,dataset - shared blank nodes,Blank nodes shared in default and named graph,0,,,TRUE,TRUE
+test073,dataset - referencing graph name,Default graph with blank node shared with graph name,0,,,TRUE,TRUE
+test074,poison - Clique Graph,A 10-node Clique of blank node resources all inter-related.,40,,,RDFC10NegativeEvalTest,FALSE
+test075,poison – evil (1),A poison graph which is computable given defined limits. (Uses SHA-384),39,,SHA384,TRUE,TRUE
```
yamdan, line 64:
```suggestion
test075,blank node - diamond (uses SHA-384),Same as test020 except for using SHA-384,0,,SHA384,TRUE,TRUE
```

yamdan (changes requested):
Thank you @gkellogg!
I agree with [David's comment](https://github.com/w3c/rdf-canon/pull/161/files#r1311015903) so I tried to prepare a SHA-384 version of `test020` here.
The original poison graph test is actually not appropriate for the mapping test because it can have multiple issued identifier maps depending on the runtime environment (due to its multiple automorphisms)."

spec/index.html
```diff
@@ -372,9 +372,9 @@ <h3>Terms defined by this specification</h3>
       <dt><dfn>hash algorithm</dfn></dt>
       <dd>The default hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256.
         <p>Implementations can be written to parameterize the
-          hash algorithm without any other changes. However, using a different
+          hash algorithm without any other changes. Using a different
           hash algorithm is expected to generate different output from
```
TallTed, line 376:
```suggestion
          hash algorithm is expected to generate different output than
```

spec/index.html
gkellogg: "is expected to" => "will generally" (it might produce the same output). What more would you like this to say?

tests/manifest.jsonld
gkellogg: I think the poison graphs can come in degrees, leading to an exponential explosion in computational/memory resources. the tests which pass are the same kind of tests as those that blow up, just of a lower degree.

But, as @yamdan mentions, the use of this specific tests is inappropriate for also testing the blank node map output, as it is ambiguous, so I'll change to another. The idea was to use an existing test that would generate different outputs depending on the hash algorithm.

spec/index.html
```diff
@@ -2892,6 +2892,7 @@ <h3>Insecure Hash Algorithms</h3>
       and implementations of it can be parameterized to use a different
       hash function, without the need to make any changes to the
       canonicalization algorithm itself.
+      However, using a different hash algorithm will generally lead to differernt results.
```
TallTed, line 2895:
```suggestion
      However, using a different hash algorithm will generally lead to different results.
```

spec/index.html
```diff
@@ -370,10 +370,10 @@ <h3>Terms defined by this specification</h3>
       <dt><dfn>hash</dfn></dt>
       <dd>The lowercase, hexadecimal representation of a message digest.</dd>
       <dt><dfn>hash algorithm</dfn></dt>
-      <dd>The default hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256.
+      <dd>The default hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256 [[FIPS-180-4]].
         <p>Implementations can be written to parameterize the
           hash algorithm without any other changes. Using a different
-          hash algorithm is expected to generate different output from
+          hash algorithm will generally generate different output than
```
TallTed, line 376:
```suggestion
          hash algorithm will generally result in different output than
```

TallTed (changes requested):
Minor wordsmithing

yamdan (approved):
Thank you for updating. I am still concerned that an empty `vocab.jsonld` file remains in the root, but other than that, I think it is good.

vocab.jsonld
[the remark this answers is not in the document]
gkellogg: I think it was generated accidentally by running mk_vocab in the wrong directory. I removed it in f98562a1.

davidlehn (approved):

spec/index.html
davidlehn: I suppose it's fine.  I wanted to make sure no one got tripped up on the issue that this is an important input parameter.

PR #167, opened 2023-09-07
Update N-Quads canonical form

### LINKED ISSUE #165: \u vs \U
"Characters in the range from U+0000 to U+001F and U+007F (DEL) that are not represented using [ECHAR](https://www.w3.org/TR/n-quads/#grammar-production-ECHAR) MUST be represented by [UCHAR](https://www.w3.org/TR/n-quads/#grammar-production-UCHAR)."

It appears to me that U+1 can be written as either \u0001 or \U00000001.

### LINKED ISSUE #166: serializing surrogate code points
It appears that surrogate code points may be allowed in the lexical form of RDF literals.  If they are, then they cannot be serialized in UTF-8, so they would have to be escaped.

Similarly for integer values in lexical forms that are not code points at all, if such are allowed.

I suggest that the canonicalization algorithm be written to work correctly for unusual elements of lexical forms, including both surrogate code points and non-code points.  This will provide some amount of future-proofing and defense against unusual elements of lexical forms.

gkellogg:
… to use code point representation from the current N-Quads spec, require that characters outside of the XML Char production are encoded via UCHAR, and that the shortest possible encoding is used for UCHAR.

Fixes #165. Fixes #166.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/167.html" title="Last updated on Sep 10, 2023, 8:23 PM UTC (5861109)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/167/5bd72cd...5861109.html" title="Last updated on Sep 10, 2023, 8:23 PM UTC (5861109)">Diff</a>


davidlehn:
- Would the I-JSON (https://www.rfc-editor.org/rfc/rfc7493.html#section-2.1) language help with any clarification here on the surrogate code points?  Or is what's written here good enough?  (I'm no expert on these issues.)
- Needs tests.
- I have a concern about the potential performance issues adding up.  For general use, some stage of processing will require an updated or additional scan and replace to ensure shortened chars.  That is unfortunate.

gkellogg:
I-JSON is well out of the set of the core that RDF is based on, which really comes from HTML. 

The added description of what is allowed in literal really already matches what N-Quads can represent, but this adds clarification based on concerns raised by @pfps. It pretty much represents what N-Quads serializers should already do, but the Char production details are somewhat hidden in the EBNF spec. Same for shortest way to represent UCHAR. So, net should be no change.

A couple of tests are useful, but due to inherent limitations of N-Quads, it will probably be limited to representing \U sequences as \u, where there is ambiguity.

davidlehn:
I suppose this is another case where it's recommended that implementations should be also be passing an N-Quads test suite.  Simplified N-Quads code is fine until you hit all the weird/questionable/bad data and edge cases.

Without making the test suite too complex it does seem like a test case or two could be written.

spec/index.html
```diff
@@ -3531,30 +3531,35 @@ <h2>A Canonical form of N-Quads</h2>
...
+          <code>CR</code> (carriage return, code point <code class="codepoint">U+000D</code>),
+          <code>&quot;</code> (quotation mark, code point <code class="codepoint">U+0022</code>), and
+          <code>\</code> (backslash, code point <code class="codepoint">U+005C</code>)
+          MUST be encoded using <code><a data-cite="N-QUADS#grammar-production-ECHAR">ECHAR</a></code>.</li>
+        <li>Characters in the range from <code class="codepoint">U+0000</code> to <code class="codepoint">U+001F</code>,
+          <code>DEL</code> (delete, code point <code class="codepoint">U+007F</code>),
+          and characters not matching the <a data-cite="XML11#charsets">Char</a> production from [[XML11]]
+          that are not represented using <code><a data-cite="N-QUADS#grammar-production-ECHAR">ECHAR</a></code>
+          MUST be represented by <code><a data-cite="N-QUADS#grammar-production-UCHAR">UCHAR</a></code>
+          using the shortest possible encoding.</li>
```
yamdan, line 3557:
```suggestion
        <li>Characters in the range from <code class="codepoint">U+0000</code> to <code class="codepoint">U+0007</code>,
          <code>VT</code> (vertical tab, code point <code class="codepoint">U+000B</code>),
          characters in the range from <code class="codepoint">U+000E</code> to <code class="codepoint">U+001F</code>,
          <code>DEL</code> (delete, code point <code class="codepoint">U+007F</code>),
          and characters not matching the <a data-cite="XML11#charsets">Char</a> production from [[XML11]]
          MUST be represented by <code><a data-cite="N-QUADS#grammar-production-UCHAR">UCHAR</a></code>
          using a lowercase <code>\u</code> with 4 <code><a data-cite="N-QUADS#grammar-production-HEX">HEX</a></code>es.</li>
```
I've enumerated the targets of the UCHAR encoding more explicitly, and
considering that only 4-HEX UCHAR encoding appears here, I've narrowed down the expression to be specific to that.
(It's my understanding that characters not matching the XML Char production should correspond to `U+0000`, `U+D800`-`U+DFFF`, `U+FFFE`, and `U+FFFF`, all of which can be encoded using 4 HEXes.)

yamdan (commented):
Thank you, I made a suggestion in hopes of providing more clarity related to UCHARs.

yamdan:
The current test case (test060) already covered 4-HEX UCHAR output (i.e., `\u0000`-`\u0007`, `\u000B`, `\u000E`-`\u001F`, and `\u007F`), so it seems to address issue #165 (`\u` vs `\U`).
On the other hand, for issue #166 (surrogate code points), I haven't been able to confirm that `U+D800`-`U+DFFF`, `U+FFFE`, and `U+FFFF` are encoded as `\uXXXX` in the canonical form of output. However, many N-Quads parsers (or at least the one I use) treat these characters as invalid and return errors. Thus, representing these characters in the test input (test060-in.nq) seems challenging.

spec/index.html
gkellogg: I think this is missing U+000C `VT` (form feed). I'm not sure that spelling these out adds all that much. Note that XML 1.1 [Char](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Char) production also includes `[#x10000-#x10FFFF]`, which are represented as `\U00010000` through `\U0010FFFF`. It allows pretty much everything but surrogates. UChar doesn't include U+0000 which is explicitly included above.

The "shortest possible encoding" is intended to infer either `\uxxxx` or `\U00xxxxxx` implicitly. The [UCHAR](https://www.w3.org/TR/n-quads/#grammar-production-UCHAR) production should be adequate documentation for what is meant here.

gkellogg:
> The current test case (test060) already covered 4-HEX UCHAR output (i.e., `\u0000`-`\u0007`, `\u000B`, `\u000E`-`\u001F`, and `\u007F`), so it seems to address issue #165 (`\u` vs `\U`).
> On the other hand, for issue #166 (surrogate code points), I haven't been able to confirm that `U+D800`-`U+DFFF`, `U+FFFE`, and `U+FFFF` are encoded as `\uXXXX` in the canonical form of output. However, many N-Quads parsers (or at least the one I use) treat these characters as invalid and return errors. Thus, representing these characters in the test input (test060-in.nq) seems challenging.

We're dealing with RDF 1.1 here, so we're somewhat limited in what we can do. RDF 1.1 sort of allows surrogates, although note in the main concrete syntaxes (other than through UCHAR). https://github.com/w3c/rdf-concepts/pull/59 (in its current form) ads a restriction that code points must be scalar values, which exclude surrogates.

iherman:
I am a little bit worried about the direction we are taking here (unless I misunderstand something). All these issues should _not_ be a matter for the RCH WG but the matter of the RDF WG, in particular, what the RDF standard says about a canonical representation of literals. Very practical matter: an RCH implementation should be in the position of 100% relying on the underlying RDF environment (Jena, RDFLib, rdfjs, whatever) to do the right thing. It should not be a matter of the RCH spec and/or implementation to mess around with the character representation.

Should we discuss this next Monday at the F2F?

yamdan:
I believe all these issues will be resolved in N-Quads 1.2. However, at this moment, the RCH spec only supports N-Quads 1.1, so we have to address them in Appendix A, within the scope necessary for canonicalization, as a temporary necessity.

spec/index.html
TallTed: Lightly tweaking the end of @yamdan's suggestion
```suggestion
        <li>Characters in the range from <code class="codepoint">U+0000</code> to <code class="codepoint">U+0007</code>,
          <code>VT</code> (vertical tab, code point <code class="codepoint">U+000B</code>),
          characters in the range from <code class="codepoint">U+000E</code> to <code class="codepoint">U+001F</code>,
          <code>DEL</code> (delete, code point <code class="codepoint">U+007F</code>),
          and characters not matching the <a data-cite="XML11#charsets">Char</a> production from [[XML11]]
          MUST be represented by <code><a data-cite="N-QUADS#grammar-production-UCHAR">UCHAR</a></code>
          using a lowercase <code>\u</code> with 4 <code><a data-cite="N-QUADS#grammar-production-HEX">HEX</a></code>es
          for characters in the range from <code class="codepoint">U+0000</code> to <code class="codepoint">U+FFFF</code>
          and using an uppercase <code>\U</code> with 8 <code><a data-cite="N-QUADS#grammar-production-HEX">HEX</a></code>es
          for characters in the range from <code class="codepoint">U+00010000</code> to <code class="codepoint">U+FFFFFFFF</code>.</li>

```

yamdan: Thanks! However, I still believe that we don't need to introduce "\U with 8 HEXes" in this context.

If I recall correctly, our goal is to convey that "characters not matching the Char production from XML11 MUST be represented by UCHAR."
Given that the "[Char](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Char) production from XML11" is defined as:

```
Char	   ::=   	[#x1-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
```

"Characters not matching the Char production" should equal to `#x0 | [#xD800-#xDFFF] | [#xFFFE-#xFFFF]`, all of which can be represented within 4 HEXes.
We don't need to consider values like `[#x110000-...]` as the Unicode set is defined from `[#x0-#x10FFFF]`.

I'm not a Unicode expert, so please correct me if I'm mistaken in any way.

TallTed: My thought was, in part, future proofing. Also, I think it's problematic to explicitly talk about the lowercase `\u` range and omit any mention of the uppercase `\U` range.

_If_ the largest Unicode code point will _always_ be lower than `#xFFFFFFFF`, then sure, the high point of what I wrote could be lowered to whatever that highest point will be ... which you say is `#x10FFFF`, so --
```
using an uppercase <code>\U</code> with 8 
<code><a data-cite="N-QUADS#grammar-production-HEX">HEX</a></code>es for 
characters in the range from <code class="codepoint">U+00010000</code> to 
<code class="codepoint">U+0010FFFF</code>.</li>
```
-- perhaps trimming from 8 to 6 `HEX`es, if omitting insignificant `0`s is always legal.

But I've been awake for far too long, so I may be mis-thinking something. :-)

gkellogg: Yes, I was too hasty about the high-range of Unicode (danger of late-night replies). There would not be a case for using the `\U00XXXXXX` form given the requirement for using the native representation.

> I believe U+000C should be encoded as ECHAR rather than UCHAR.

The set of characters represented by ECHAR `"\" [tbnrf"']` is long established, and I don't think there's much of a desire to add to it at this point.

Regarding future-proofing, the specs reference a specific version of Unicode, which explicitly defines the different [character blocks](https://en.wikipedia.org/wiki/Unicode_block#List_of_blocks). Anything outside of this range would not be valid from XML 1.1, which is our basis, so adding provision for an encoding that can't be used isn't too useful.

gkellogg:
> I believe all these issues will be resolved in N-Quads 1.2. However, at this moment, the RCH spec only supports N-Quads 1.1, so we have to address them in Appendix A, within the scope necessary for canonicalization, as a temporary necessity.

Yes, I think we'll want to discuss this. Although we're trying to keep the language for N-Quads symmetric here with RDF 1.2, RDF 1.1 technically allows Unicode surrogates to be represented using UCHAR, and RDF 1.2 likely won't. But, I think having explicit tests using such surrogate UCHAR representations may prove to be a problem, based on @yamdan's experience. It's considered a bug that it is supported in earlier versions of RDF, and I don't think we want to add them to tests at this point.

pfps:
One can read RDF 1.1 Concepts to say that a lexical form is a sequence of 32-bit integers.  This can be serialized using the \U escape and could be useful in a datatype that is a sequence of 32-bit integers.  I don't expect that any current RDF implementation supports this, but these sorts of weird edge cases should be supported if at all possible in canonical forms.

spec/index.html
```diff
@@ -3531,30 +3531,35 @@ <h2>A Canonical form of N-Quads</h2>
...
       <code>object</code>,
       and <code>graphLabel</code>,
-      each of which MUST be a single space (<code>U+0020</code>).</li>
+      each of which MUST be a single space (code point <code class="codepoint">U+0020</code>).</li>
     <li><a data-cite="RDF11-CONCEPTS#dfn-literal">Literals</a> with the
       datatype <code>http://www.w3.org/2001/XMLSchema#string</code>
       MUST NOT use the datatype IRI part of the <a data-cite="N-QUADS#grammar-production-literal">literal</a>,
       and are represented using only <a data-cite="N-QUADS#grammar-production-STRING_LITERAL_QUOTE">STRING_LITERAL_QUOTE</a>.
     </li>
     <li><code><a data-cite="N-QUADS#grammar-production-HEX">HEX</a></code> MUST use only uppercase letters (<code>[A-F]</code>).</li>
```
gkellogg, line 3540:
```suggestion
    <li><code><a data-cite="N-QUADS#grammar-production-HEX">HEX</a></code> MUST use only digits `[0-9]` and uppercase letters (<code>[A-F]</code>).</li>
```

spec/index.html
```diff
@@ -3531,30 +3531,35 @@ <h2>A Canonical form of N-Quads</h2>
...
+          <code>&quot;</code> (quotation mark, code point <code class="codepoint">U+0022</code>), and
+          <code>\</code> (backslash, code point <code class="codepoint">U+005C</code>)
+          MUST be encoded using <code><a data-cite="N-QUADS#grammar-production-ECHAR">ECHAR</a></code>.</li>
+        <li>Characters in the range from <code class="codepoint">U+0000</code> to <code class="codepoint">U+001F</code>,
+          <code>DEL</code> (delete, code point <code class="codepoint">U+007F</code>),
+          and characters not matching the <a data-cite="XML11#charsets">Char</a> production from [[XML11]]
+          that are not represented using <code><a data-cite="N-QUADS#grammar-production-ECHAR">ECHAR</a></code>
+          MUST be represented by <code><a data-cite="N-QUADS#grammar-production-UCHAR">UCHAR</a></code>
+          using the shortest possible encoding.</li>
+        <li>All other characters MUST be represented by their native [[UNICODE]] representation.</li>
```
gkellogg, line 3558:
```suggestion
        <li>All characters not required to be represented by
          <code><a data-cite="N-QUADS#grammar-production-ECHAR">ECHAR</a></code> or
          <code><a data-cite="N-QUADS#grammar-production-UCHAR">UCHAR</a></code>
          MUST be represented by their native [[UNICODE]] representation.</li>
```

spec/index.html
```diff
@@ -3531,30 +3531,39 @@ <h2>A Canonical form of N-Quads</h2>
...
-      <code>U+005C</code> (<code title="BACKSLASH">\</code>)
-      MUST be encoded using <code><a data-cite="N-QUADS#grammar-production-ECHAR">ECHAR</a></code>.
-      Characters in the range from <code>U+0000</code> to <code>U+001F</code>
-      and <code>U+007F</code>  (<code title="delete"><sub>DEL</sub></code>)
-      that are not represented using <code><a data-cite="N-QUADS#grammar-production-ECHAR">ECHAR</a></code>
-      MUST be represented by <code><a data-cite="N-QUADS#grammar-production-UCHAR">UCHAR</a></code>.
-      All other characters MUST be represented by their native [[UNICODE]] representation.</li>
-    <li>The token <code><a data-cite="N-QUADS#grammar-production-EOL">EOL</a></code> MUST be a single <code>U+000A</code>.</li>
-    <li>The final <code><a data-cite="N-QUADS#grammar-production-EOL">EOL</a></code> MUST be provided.</li>
+    <li><code><a data-cite="N-QUADS#grammar-production-HEX">HEX</a></code> MUST use only digits `[0-9]` and uppercase letters (<code>[A-F]</code>).</li>
```
yamdan, line 3540:
```suggestion
    <li><code><a data-cite="N-QUADS#grammar-production-HEX">HEX</a></code> MUST use only digits (<code>[0-9]</code>) and uppercase letters (<code>[A-F]</code>).</li>
```
to align with the other parts like the subsequent `(<code>[A-F]</code>)`

yamdan (approved):
Thank you @gkellogg for updating. I approved with a minor editorial suggestion.

> One can read RDF 1.1 Concepts to say that a lexical form is a sequence of 32-bit integers.

I might be overlooking something, but it seems to me that the [lexical form in RDF 1.1 Concepts](https://www.w3.org/TR/rdf11-concepts/#dfn-lexical-form) says it is a Unicode string rather than a sequence of 32-bit integers.

pfps:
> Thank you @gkellogg for updating. I approved with a minor editorial suggestion.
> 
> > One can read RDF 1.1 Concepts to say that a lexical form is a sequence of 32-bit integers.
> 
> I might be overlooking something, but it seems to me that the [lexical form in RDF 1.1 Concepts](https://www.w3.org/TR/rdf11-concepts/#dfn-lexical-form) says it is a Unicode string rather than a sequence of 32-bit integers.

From Unicode 15.0, Section 2.7: "A Unicode string data type is simply an ordered sequence of code units. Thus a Unicode 8-bit string is an ordered sequence of 8-bit code units, a Unicode 16-bit string is an ordered sequence of 16-bit code units, and a Unicode 32-bit string is an ordered sequence of 32-bit code units."   From Section 2.5: "Actual implementations in computer systems represent integers in specific code units of
particular size—usually 8-bit (= byte), 16-bit, or 32-bit."

yamdan:
Yes, but considering the following description within the same Unicode document, it seems to me that we don't need to assume the existence of code points larger than `U+10FFFF`.
> [From Unicode 15.0, Section 2.4] "In the Unicode Standard, the codespace consists of the integers from 0 to 10FFFF_{16}, comprising 1,114,112 code points available for assigning the repertoire of abstract characters."

pfps:
The problem is that lexical forms may not have to be a sequence of Unicode code points in RDF 1.1.   Is `"\UABCD1234\UABCD1235"^^ex:32bitbinary .` valid RDF under a loose reading of Unicode string?  It seems so to me.

gkellogg:
I think N-Triples and Turtle make it clear that the UCODE sequence is for specifying Unicode characters/code points, not arbitrary integer values. I don't believe that `"\UABCD1234\UABCD1235"^^ex:32bitbinary` is valid RDF.

An RDF Literal has a lexical form, which is a Unicode string. XSD provides for binary forms which use Unicode to represent arbitrary data (properly encoded). Any other datatype would need to make sure that the lexical form is Unicode.

pfps:
> An RDF Literal has a lexical form, which is a Unicode string. 

And here is the problem.   A Unicode string is a sequence of code units, not scalar values or code points.  I believe that it was a mistake to use "Unicode string" in RDF Concepts (the original version), but that's what has existed for 20 years.

TallTed (approved):

PR #173, opened 2023-09-11
Make the use of a parameter for selecting the hash algorithm a MUST.

### LINKED ISSUE #169: Make the possibility of changing the hash function a SHOULD?
We know that RDFC has been criticized on the ground of using SHA-256. Whether this is justified or not, it is not for me to say; I am not an expert on hash functions. The fact remains that it is a point of "vulnerability".

Our current answer is in the (terminology section)[https://www.w3.org/TR/rdf-canon/#dfn-hash-algorithm]:

> The default hash algorithm used by [RDFC-1.0](https://www.w3.org/TR/rdf-canon/#dfn-rdfc-1-0), namely, SHA-256 [[FIPS-180-4](https://www.w3.org/TR/rdf-canon/#bib-fips-180-4)].
>
> Implementations can be written to parameterize the hash algorithm without any other changes. Using a different hash algorithm will generally result in different output than using the default.

In the [specification itself](https://www.w3.org/TR/rdf-canon/#canon-overview), it only says:

> By default, [RDFC-1.0](https://www.w3.org/TR/rdf-canon/#dfn-rdfc-1-0) uses the SHA-256 hash algorithm [[FIPS-180-4](https://www.w3.org/TR/rdf-canon/#bib-fips-180-4)].

My feeling is that these statements are weak in view of the context. In my view, we should say somewhere in the spec (probably in [§4.1](https://www.w3.org/TR/rdf-canon/#canon-overview)) that

- Implementation _SHOULD_ be written in a way that the hash algorithm can be parametrized
- If such a parametrization is built in the implementation, at the minimum SHA-384 _MUST_ be provided as an alternative.

(And we can say something about future versions, which may mandate other alternatives.)

Actually, I would not be shocked to see _SHOULD_ be a _MUST_.

Note that, at this stage, this is an editorial change only, that does not jeopardize the fate of the spec: we have two tests which check the usage of SHA-384, and we already have three implementations that pass these tests. I.e., we are ready to go to REC even with _MUST_.

gkellogg:
Fixes #169.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/pull/173.html" title="Last updated on Sep 11, 2023, 10:04 AM UTC (b6eacd7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/173/97f0f4d...b6eacd7.html" title="Last updated on Sep 11, 2023, 10:04 AM UTC (b6eacd7)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -369,8 +369,8 @@
       <dd>The lowercase, hexadecimal representation of a message digest.</dd>
       <dt><dfn>hash algorithm</dfn></dt>
       <dd>The default hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256 [[FIPS-180-4]].
-        <p>Implementations MUST support a parameter to define the <a>hash algorithm</a>
-          and MUST support SHA-256 and SHA-384 [[FIPS-180-4]]
+        <p>Implementations MUST support a parameter to define the <a>hash algorithm</a>,
+          MUST support SHA-256 and SHA-384 [[FIPS-180-4]],
           and SHOULD support other hash algorithms.
           Using a different hash algorithm will generally result in different output than
           using the default.</p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -280,8 +280,6 @@
       it may be possible to correlate the original blank node identifiers
       used within that N-Quads document with those issued in the
       <a>canonicalized dataset</a>.</p>
-
-    <div class="issue" data-number="19"></div>
   </section>
 
   <section id="how-to-read">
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -371,9 +369,10 @@ <h3>Terms defined by this specification</h3>
       <dd>The lowercase, hexadecimal representation of a message digest.</dd>
       <dt><dfn>hash algorithm</dfn></dt>
       <dd>The default hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256 [[FIPS-180-4]].
-        <p>Implementations can be written to parameterize the
-          hash algorithm without any other changes. Using a different
-          hash algorithm will generally result in different output than
+        <p>Implementations MUST support a parameter to define the <a>hash algorithm</a>
+          and MUST support SHA-256 and SHA-384 [[FIPS-180-4]]
+          and SHOULD support other hash algorithms.
```
TallTed, line 374:
```suggestion
        <p>Implementations MUST support a parameter to define the <a>hash algorithm</a>,
          MUST support SHA-256 and SHA-384 [[FIPS-180-4]],
          and SHOULD support other hash algorithms.
```

TallTed (changes requested):

iherman (approved):

spec/index.html
```diff
@@ -371,9 +369,10 @@ <h3>Terms defined by this specification</h3>
       <dd>The lowercase, hexadecimal representation of a message digest.</dd>
       <dt><dfn>hash algorithm</dfn></dt>
       <dd>The default hash algorithm used by <a>RDFC-1.0</a>, namely, SHA-256 [[FIPS-180-4]].
-        <p>Implementations can be written to parameterize the
-          hash algorithm without any other changes. Using a different
-          hash algorithm will generally result in different output than
+        <p>Implementations MUST support a parameter to define the <a>hash algorithm</a>,
+          MUST support SHA-256 and SHA-384 [[FIPS-180-4]],
+          and SHOULD support other hash algorithms.
```
gkellogg, line 374: and SHOULD support the ability to add other hash algorithms.

peacekeeper (approved):

PR #178, opened 2023-09-30
Add some caveats about changing the default hash algorithm

### LINKED ISSUE #176: Issues with parameterized hashing algorithms used internally
See Sebastian's [email of 2023-09-12](https://lists.w3.org/Archives/Public/public-rch-wg/2023Sep/0010.html) and subsequent disucssion

gkellogg:
and advice on keeping the algorithm used along with any results.

(Note, when this PR is created, spec-prod is having some issues resolving aspects of the document, so the PR Preview link might show errors. This [GitHack](https://raw.githack.com/w3c/rdf-canon/hash-algo-clarifications/spec/index.html) version may provide better results).

Fixes #176.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -270,7 +270,9 @@
       As a result, a graph signature can be obtained by hashing a canonical serialization
       of the resulting <a>canonicalized dataset</a>,
       allowing for the isomorphism and digital signing use cases.
-      As blank node identifiers can be stable even with other changes to a graph (dataset),
+      This algorithm does not define such a graph signature.</p>
+
+    <p>As blank node identifiers can be stable even with other changes to a graph (dataset),
       in some cases it is possible to compute the difference between two graphs (datasets),
       for example if changes are made only to ground triples,
       or if new blank nodes are introduced which do not create an automorphic confusion
@@ -281,6 +283,15 @@
       it may be possible to correlate the original blank node identifiers
       used within that N-Quads document with those issued in the
       <a>canonicalized dataset</a>.</p>
+
+    <p class="note">Although alternative <a>hash algorithms</a> may be used
+      with this specification,
+      applications should carfully weigh the advantages
+      and disadvantates of using an alternative hash function.
+      In particular, any representation of the <a>canonical n-quads form</a>
+      or <a>issued identifiers map</a>
+      without identifying the associated hash algorithm.
+    </p>
   </section>
 
   <section id="how-to-read">
@@ -375,6 +386,9 @@
           and SHOULD support the ability to specify other hash algorithms.
           Using a different hash algorithm will generally result in different output than
           using the default.</p>
+
+        <p class="note">There is no expectation that the default hash function
+          is also used to create a hash signature of the canonical N-Quads result.</p>
       </dd>
       <dt><dfn>mention</dfn></dt>
       <dd>
@@ -2881,8 +2895,15 @@
       and implementations of it can be parameterized to use a different
       hash function, without the need to make any changes to the
       canonicalization algorithm itself.
-      However, using a different hash algorithm will generally lead to different results.
+      However, using a different hash algorithm will generally lead to different results;
+      applications making use of this specification should carfully weigh the advantages
+      and disadvantates of using an alternative hash function.
     </p>
+
+    <p class="note">The possible implications of the default hash algorithm
+      becoming insecure are mitegated by that fact that no internal hash
+      values are revealed, and the algorithms are designed to cope
+      with hash collisions.</p>
   </section>
 
 </section>


spec/index.html
```diff
@@ -281,6 +283,15 @@ <h2>Uses of Dataset Canonicalization</h2>
       it may be possible to correlate the original blank node identifiers
       used within that N-Quads document with those issued in the
       <a>canonicalized dataset</a>.</p>
+
+    <p class="note">Although alternative <a>hash algorithms</a> may be used
+      with this specification,
+      applications should carfully weigh the advantages
```
iherman, line 289:
```suggestion
      applications should carefully weigh the advantages
```

spec/index.html
```diff
@@ -281,6 +283,15 @@ <h2>Uses of Dataset Canonicalization</h2>
...
+      and disadvantates of using an alternative hash function.
+      In particular, any representation of the <a>canonical n-quads form</a>
+      or <a>issued identifiers map</a>
+      without identifying the associated hash algorithm.
```
iherman, line 293:
```suggestion
      This is the case, in particular, for any representation of the <a>canonical n-quads form</a>
      or <a>issued identifiers map</a>
      without identifying the associated hash algorithm.
```

iherman (approved):

spec/index.html
```diff
@@ -281,6 +283,15 @@ <h2>Uses of Dataset Canonicalization</h2>
...
       used within that N-Quads document with those issued in the
       <a>canonicalized dataset</a>.</p>
+
+    <p class="note">Although alternative <a>hash algorithms</a> may be used
+      with this specification,
+      applications should carefully weigh the advantages
+      and disadvantates of using an alternative hash function.
+      This is the case, in particular, for any representation of the <a>canonical n-quads form</a>
+      or <a>issued identifiers map</a>
+      without identifying the associated hash algorithm.
```
dlongley, line 293:
```suggestion
      that does not identify the associated hash algorithm. Any use case
      that requires reproduction of the same output is expected to
      unequivocally express or communicate the hash algorithm.
```

spec/index.html
```diff
@@ -375,6 +386,9 @@ <h3>Terms defined by this specification</h3>
           and SHOULD support the ability to specify other hash algorithms.
           Using a different hash algorithm will generally result in different output than
           using the default.</p>
+
+        <p class="note">There is no expectation that the default hash function
+          is also used to create a hash signature of the canonical N-Quads result.</p>
```
dlongley, line 391: For consistency ("function" => "algorithm") and minor wording adjustment:

```suggestion
        <p class="note">There is no expectation that the default hash algorithm
          will also be used by any application creating a hash digest of the
          canonical N-Quads result.</p>
```

dlongley (approved):
A few suggestions. Thanks!

spec/index.html
```diff
@@ -2881,8 +2898,15 @@ <h3>Insecure Hash Algorithms</h3>
...
-      However, using a different hash algorithm will generally lead to different results.
+      However, using a different hash algorithm will generally lead to different results;
+      applications making use of this specification should carfully weigh the advantages
+      and disadvantates of using an alternative hash function.
     </p>
+
+    <p class="note">The possible implications of the default hash algorithm
+      becoming insecure are mitegated by that fact that no internal hash
+      values are revealed, and the algorithms are designed to cope
+      with hash collisions.</p>
```
yamdan, line 2909: It seems to me that "to cope with hash collisions" might be a bit of an overstatement because a hash collision leads to a kind of uncertainty to the output of the algorithm.
As I mentioned in the mailing list, I believe that a hash collision leads to *ties* in for-loops of [step 4](https://www.w3.org/TR/rdf-canon/#ca.4), [step 5](https://www.w3.org/TR/rdf-canon/#ca.5), and [step 5.3](https://www.w3.org/TR/rdf-canon/#ca.5.3) in the Canonicalization Algorithm. These ties can't be deterministically resolved, making the outcome indeterminate. 
To claim it 'copes with collisions', I believe we'd need to incorporate the existence check for ties into the algorithm. However, this might complicate the algorithm and may not be easily accepted so I personally suggest just omitting this phrase here and not to modify the algorithm itself.

yamdan (commented):
Thank you @gkellogg! I just added a few comments.

spec/index.html
gkellogg: From what I can tell, there are different kinds of hash collisions:

* Two strings may result in the same hash value.
* Two quads may have the same first-order hash.
* A graph may have homomorphisms (n-degree or poison graph).

My statement is intended to consider  the first case, algorithms such as CRC, MD5, and SHA-1 are more likely to collide on a given input than higher order SHA variants. My understanding that a collision using SHA-256 is already quote low. [One answer on Quara](https://www.quora.com/Is-there-a-collision-in-SHA256#:~:text=Re%20“Is%20there%20a%20collision,256%20%3D%201.16x10%5E77) puts this in 1 chance in 2^256 (presumably with SHA-384 being 1 chance in 2^384). In my view, these are already so low as to be extremely unlikely and effectively of equivalent risk.

The problems @yamdan highlights are in hash 1-degree, and hash n-degree results, which don't relate to the quality of the SHA-256 hash, but in underlying graph homomorphisms which exist independent of any given hash function.

The "copes with collisions" claim is that, in the unlikely event that two distinct quads result in the same hash without having the same string representation, this would promote the problem to the hash n-degrees use case, which could have been avoided.

I think removing a statement entirely doesn't help in conveying the different types of hash collisions involved, and the relative risk of SHA-256 compared with SHA-384.

yamdan: > The "copes with collisions" claim is that, in the unlikely event that two distinct quads result in the same hash without having the same string representation, this would promote the problem to the hash n-degrees use case, which could have been avoided.

Thank you for the explanation. I indeed overlooked that. As you mentioned, in the loop of [step 4](https://www.w3.org/TR/rdf-canon/#ca.4) and [step 5](https://www.w3.org/TR/rdf-canon/#ca.5), we have already anticipated the possibility of collisions with the 1-degree hash and have taken measures to promote it to the Hash N-Degree Quads algorithm. Thus, I believe RDFC-1.0 already *copes with hash 1-degree collisions*.

I am still curious about what would happen if we encounter unexpected n-degree hash collisions in [step 5.3](https://www.w3.org/TR/rdf-canon/#ca.5.3) and related hash collisions in [step 5 in Hash N-Degree Quads](https://www.w3.org/TR/rdf-canon/#hndq.5).

spec/index.html
```diff
@@ -281,6 +283,17 @@ <h2>Uses of Dataset Canonicalization</h2>
...
+
+    <p class="note">Although alternative <a>hash algorithms</a> might be used
+      with this specification,
+      applications ought to carefully weigh the advantages
+      and disadvantages of using an alternative hash function.
+      This is the case, in particular, for any representation of the <a>canonical n-quads form</a>
+      or <a>issued identifiers map</a>
+      that does not identify the associated hash algorithm. Any use case
+      that requires reproduction of the same output is expected to
+      unequivocally express or communicate the hash algorithm.
```
msporny, line 295:
```suggestion
      unequivocally express or communicate the internal 
      hash algorithm that was used when generating 
      the <a>canonical n-quads form</a>.
```

msporny (approved):
LGTM (with one minor nit).

spec/index.html
gkellogg: Do you have a suggestion to reword this that better reflects your concerns while still showing the robustness of SHA-256?

yamdan:
```suggestion
      values are revealed, and the algorithms are designed to cope
      with first-degree hash collisions.</p>
```
By focusing on areas we can address clearly, I believe it becomes more accurate. How about the above rewording?

spec/index.html
```diff
@@ -2881,8 +2898,15 @@ <h3>Insecure Hash Algorithms</h3>
...
       canonicalization algorithm itself.
-      However, using a different hash algorithm will generally lead to different results.
+      However, using a different hash algorithm will generally lead to different results;
+      applications making use of this specification should carefully weigh the advantages
+      and disadvantages of using an alternative hash function.
     </p>
+
+    <p class="note">The possible implications of the default hash algorithm
+      becoming insecure are mitigated by that fact that no internal hash
+      values are revealed, and the algorithms are designed to cope
```
davidlehn, line 2908: Something like this reads a better to me.  No confusion on hash vs canonicalization algorithm.  And I think this can be singular?
```suggestion
      values are revealed, and the canonicalization algorithm is designed to cope
```

yamdan (approved):
Thanks!

PR #188, opened 2023-11-28
Update rust-sophia-rdfc10.ttl  …

### LINKED ISSUE #185: Sophia EARL results not in generated report.
The Sophia implementation data (`./reports/rust-sophia-rdfc10.ttl`) results in many incorrect "untested" results in the generated EARL data, a SPARQL warning when the report generator process is run, and no results in the HTML report.

gkellogg:
… to use the Project IRI rather than the Release IRI in assertions and use the proper prefix for the EARL vocabulary.

Fixes #185.


pchampin (approved):
LGTM

PR #192, opened 2024-02-08
Added two new tests with duplicate triples

### LINKED ISSUE #191: What is the way to go with duplicate triples?
This issue came up in an issue raised on my implementation (https://github.com/iherman/rdfjs-c14n/issues/10). The question boils down to the following.

Say the input is the following:

```turtle
@prefix ex: <http://example.com/#> .
ex:one ex:two _:a .
ex:one ex:two _:a .               
```

this is definitely a valid Turtle. When I run it through my canonicalization, the canonical quads are:

```turtle
<http://example.com/#one> <http://example.com/#two> _:c14n0 .
<http://example.com/#one> <http://example.com/#two> _:c14n0 .
```

is this the correct output? Or should it be

```turtle
<http://example.com/#one> <http://example.com/#two> _:c14n0 .
```

Clearly, the hash will be different, so we have to have a clear answer.

I am tempted to say that the output is correct. Otherwise, we would have to do some preprocessing on the nquads that the algorithm receives to filter out duplicates; but that does not seem something that should be part of RDFC... 

cc @jeswr @pchampin @dlongley @gkellogg

iherman:
This PR is to fix #191.

Adds two tests: each contain two identical triples in the input (without and with an identical blank node, respectively); the parsing of the the `nq` files should make it sure that a _dataset_ is canonicalized, ie, the duplication of triples disappear.


gkellogg (approved):

dlongley (approved):

seabass-labrax (approved):

gkellogg:
Without objection, I think we should merge this and allow updated implementation reports to be submitted.

davidlehn (approved):

davidlehn:
[rdf-canonize](https://github.com/digitalbazaar/rdf-canonize) passes the tests here and had a local test for this already.  It used a variation of the tests here with subject and object blank nodes.  Should that be added here too or is it redundant?

Input:
```
_:s <https://www.example.org/p> _:o .
_:s <https://www.example.org/p> _:o .
```

Output:
```
_:c14n1 <ex:p> _:c14n0 .
```

gkellogg:
If those tests would add something not otherwise caught, then we should add them to this PR.

iherman:
> If those tests would add something not otherwise caught, then we should add them to this PR.

True. But I am not sure that tests really add something. The issue to be tested is whether the implementation correctly filters duplicate quads in an nquads representation of a graph, whether those quads/triples include (identical) blank nodes or not. The fact that there is one bnode or there are two does not change the situation imho.

yamdan (approved):
My implementation also passes the added two tests. I will update the report after the PR is merged.

pchampin (approved):

pchampin:
FTR, my implementation also passes these tests

PR #201, opened 2024-03-01
Update EARL report for RDF.ex

marcelotto:
This adds test results (all passing) for the previously not covered `RDFC10MapTest` and `RDFC10NegativeEvalTest`.

### DIFF
--- a/reports/earl.jsonld
+++ b/reports/earl.jsonld
@@ -135,7 +135,7 @@
       "name": "earl-report-0.9.0",
       "doap:created": {
         "@type": "http://www.w3.org/2001/XMLSchema#date",
-        "@value": "2024-02-29"
+        "@value": "2024-02-28"
       },
       "revision": "0.9.0"
     },
@@ -191,7 +191,7 @@
       "doapDesc": "A RDF Dataset Canonicalization processor for JavaScript",
       "language": "JavaScript",
       "release": {
-        "@id": "_:b128",
+        "@id": "_:b172",
         "revision": "4.0.1"
       }
     },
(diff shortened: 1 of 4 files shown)


dlongley (approved):

gkellogg (approved):
Great! If you could re-un the rake task to create the generated files that would help from figuring out how to get the Git Action to push the auto-generated versions into your branch.

marcelotto:
It seems the automatic report generation doesn't work again. The used GitHub action `stefanzweifel/git-auto-commit-action` does indeed not support PRs from forks as its [README](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#use-in-forks-from-public-repositories) says:

> By default, this Action will not run on Pull Requests which have been opened by forks. (This is a limitation by GitHub, not by us.)

gkellogg:
> It seems the automatic report generation doesn't work again. The used GitHub action `stefanzweifel/git-auto-commit-action` does indeed not support PRs from forks as its [README](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#use-in-forks-from-public-repositories) says:
> 
> > By default, this Action will not run on Pull Requests which have been opened by forks. (This is a limitation by GitHub, not by us.)

Indeed, probably not worth designing a new mechanism now. Future updates could always be done by simply mailing in the EARL report to the WG mailing list and one of us can create a local branch for a PR. A nice-to-have feature would be to inject a remote branch used for a PR into the local repository to make it easier to work with.

In the mean time, thanks for doing the generation.

PR #199, opened 2024-02-29
Add list of previous work that contributed to the science.

msporny:
This PR attempts to address issue #19 by adding a history of previous work that contributed to the science that this specification is based upon.

***
<a href="https://pr-preview.s3.amazonaws.com/msporny/rdf-canon/pull/199.html" title="Last updated on Mar 1, 2024, 5:26 PM UTC (56e21f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/rdf-canon/199/9cf1297...msporny:56e21f9.html" title="Last updated on Mar 1, 2024, 5:26 PM UTC (56e21f9)">Diff</a>

PR #203, opened 2024-03-02
Pruning the prior art references

iherman:
(As agreed in https://github.com/w3c/rdf-canon/pull/199#issuecomment-1973415321, creating this as a separate PR after the merge of #199.)

As argued in https://github.com/w3c/rdf-canon/pull/199#pullrequestreview-1910510063, I believe it is unnecessary to bring in non-RDF references. It is not meaningful for our readers, and it looks like an arbitrary choice from the _huge_ literature on graph theory in general.

(I understand that some of these references may have been relevant for some of the cited works, i.e., they may be, sort of, "second level" references. But the only goal of this section is to show that the spec relies on a bunch of works that have been done in the RDF community. It is not the place to go beyond that imho.)

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -3802,56 +3802,6 @@
 
   <ul>
     <li>
-      Corneil, D.G., Gotlieb, C.C.:
-      An Efficient Algorithm for Graph Isomorphism.
-      J. ACM. 17, 51–64 (1970).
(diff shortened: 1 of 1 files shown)


gkellogg (approved):

dlongley (approved):

yamdan (approved):

philarcher (approved):
No argument from me. Thanks.

peacekeeper (approved):
Looks good.

